PR #13, opened 2011-09-01
Reverted back to your original changes and merged /gravatar command.

jasonpbarrera:


Chat/Chat.js
```diff
@@ -250,12 +264,6 @@ $(function () {
 
             addMessage(user.Name + ' left ' + this.room, 'notification');
         }
-        else {
-            clearMessages();
-            $('#users li').not('#u-' + user.Id).remove();
-
-            addMessage('You have left ' + this.room, 'notification');
```
davidfowl, line 257: Nothing here should be deleted

Chat/Chat.cs
```diff
@@ -354,6 +359,40 @@ namespace SignalR.Samples.Hubs.Chat {
...
         }
 
+        private void HandleGravatar(string name, string[] parts){
+            string email = string.Join(" ", parts.Skip(1));
+
+            if (string.IsNullOrWhiteSpace(email)){
+                throw new InvalidOperationException("Email was not specified!");
+            }
+
+            if (!_users.ContainsKey(name)){
```
davidfowl: You don't need this check, EnsureUser does this.

Chat/Chat.cs
```diff
@@ -354,6 +359,40 @@ namespace SignalR.Samples.Hubs.Chat {
...
+            user.Hash = email.ToMD5();
+
+            bool inRooms = _userRooms[name].Any();
+
+            if (inRooms)
+            {
+                foreach (var room in _userRooms[name])
+                {
+                    _rooms[room].Users.Remove(name);
+                    _rooms[room].Users.Add(name);
```
davidfowl: What's the point of these 2 lines.

PR #29, opened 2011-09-09
Added cookies for user name and gravatar.

ilprexb:


Chat/ViewModels/UserViewModel.cs
```diff
@@ -3,12 +3,14 @@
...
     public class UserViewModel {
         public UserViewModel(ChatUser user) {
+            Gravatar = user.Gravatar;
             Name = user.Name;
             Hash = user.Hash;
             Id = user.Id;
             Active = user.Active;
         }
 
+        public string Gravatar { get; set; }
```
davidfowl, line 12: Why do you need this?

Chat/Models/ChatUser.cs
```diff
@@ -4,6 +4,7 @@ using System;
 namespace Chat.Models {
     public class ChatUser {
         public string Id { get; set; }
+        public string Gravatar { get; set; }
```
davidfowl: Why do you need this?

ilprexb: For both emails, you probably don't need that in there. I just added it in case you were going to track this stuff in the future on the server side as opposed to on the client side. Also, I didn't change the expire time on the cookies in Chat.js, I didn't want to step on your project and don't know how long you wanted those set for so I just left them.> Date: Fri, 9 Sep 2011 10:47:23 -0700

davidfowl:
You're storing the cookie correctly but not using it on the server correctly. Right now we store the user id as a cookie on the client, that works as long as the appdomain isn't reset. When the app domain is reset we still want to set the user's state (nick and gravatar) without needing the userid cookie. I'm not sure if this pull request satisfies that.

ilprexb:
Hi Dave, My assumption was that the Context.ClientId will change, but the users name won't. If the userid cookie doesnt match you still have the user name and gravatar cookies. Maybe I misunderstood the internal workings of the Context.ClientId, was late, lol. Sorry about that. I guess clearing the cookie if the userid doesn't match probably isnt required either in that case, but at the same time I wasn't sure how often the recycling was happening, or how long you wanted to keep them for. -Joe

davidfowl:
No problem, given that knowledge you can re-implement the feature.

ilprexb:
I posted another commit a few hours ago. Thanks David.

PR #30, opened 2011-09-10
Autocomplete and gravatar

Buildstarted:
Added autocomplete="off" to chat input to prevent dropdown.

Added ToLowerInvariant() to the user.email for gravatar per site recommendation. http://en.gravatar.com/site/implement/hash/


davidfowl:
Hmm dunno why this didnt close

PR #31, opened 2011-09-10
Reimplementation of user name and gravatar cookies.

ilprexb:


davidfowl:
Did you update your fork before sending the pull request?

Chat/Chat.cs
```diff
@@ -41,16 +41,23 @@ namespace SignalR.Samples.Hubs.Chat {
             Caller.version = typeof(Chat).Assembly.GetName().Version.ToString();
 
             // Check the user id cookie
-            HttpCookie cookie = Context.Cookies["userid"];
+            HttpCookie cookie = Context.Cookies["username"];
             if (cookie == null) {
                 return false;
             }
 
-            ChatUser user = _db.Users.FirstOrDefault(u => u.Id == cookie.Value);
```
davidfowl: This is wrong. We should only be looking at the name if the id fails.

Chat/Chat.cs
```diff
@@ -41,16 +41,23 @@ namespace SignalR.Samples.Hubs.Chat {
             Caller.version = typeof(Chat).Assembly.GetName().Version.ToString();
 
             // Check the user id cookie
-            HttpCookie cookie = Context.Cookies["userid"];
```
davidfowl: We should only be looking at the name if the id isn't there.

PR #39, opened 2011-09-30
Server reset and Browser refresh support

daanl:
Server reset and Browser refresh support

### DIFF
--- a/Chat/Chat.cs
+++ b/Chat/Chat.cs
@@ -58,11 +58,7 @@ namespace SignalR.Samples.Hubs.Chat {
             // if we could get user by id try it by name server could be resetted
             if (user == null && userNameCookie != null && !string.IsNullOrWhiteSpace(userNameCookie.Value))
             {
-                var userName = userNameCookie.Value;
-               
-                EnsureUserNameIsAvailable(userName);
-
-                user = AddUser(userName);
+                user = AddUser(userNameCookie.Value);
             }
 
             // if we have no user return false will force user to set new nick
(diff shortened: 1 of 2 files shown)


Chat/Chat.cs
```diff
@@ -43,18 +43,57 @@ namespace SignalR.Samples.Hubs.Chat {
...
 
-            ChatUser user = _db.Users.FirstOrDefault(u => u.Id == cookie.Value);
+            // if we could get user by id try it by name server could be resetted
+            if (user == null && userNameCookie != null && !string.IsNullOrWhiteSpace(userNameCookie.Value))
+            {
+                var userName = userNameCookie.Value;
 
-            // If there's no registered user, return false
-            if (user == null) {
+                var userExists = _db.Users.Where(u => u.Name.Equals(userName, StringComparison.OrdinalIgnoreCase)).Any();
```
davidfowl: You can do Any(condition) you don't need Where(condition).Any().

Chat/Chat.cs
```diff
@@ -43,18 +43,57 @@ namespace SignalR.Samples.Hubs.Chat {
...
+
+                if (userExists) {
+                    throw new InvalidOperationException(string.Format("Username {0} already taken, please choose a new one", userName));
```
davidfowl: Don't we do this somewhere else, reactor that method into something and call it from both places.

Chat/Chat.cs
```diff
@@ -588,6 +623,17 @@ namespace SignalR.Samples.Hubs.Chat {
 
             var userViewModel = new UserViewModel(user);
             Caller.addUser(userViewModel);
+
+            return user;
+        }
+
+        private ChatRoom AddRoom(string name)
```
davidfowl, line 718: Don't we do this somewhere else as well? Why not use the same code?

PR #47, opened 2011-10-04
Fix for Issue #46: Html from the PastieContentProvider doesn't scroll to bottom

robertwesterlund:
Changed from using the captureWrite jquery plugin to using the captureDocumentWrite jquery plugin, to allow scrolling to bottom on pasties.


Chat/ContentProviders/PastieContentProvider.cs
```diff
@@ -21,7 +21,18 @@ namespace SignalR.Samples.Hubs.Chat.ContentProviders {
         }
 
         public override string MediaFormatString {
-            get { return @"<script src='http://pastie.org/{0}.js'></script>"; }
+            get
+            {
+                var scriptTagId = Guid.NewGuid().ToString();
```
davidfowl, line 26: What's this for?

robertwesterlund, line 26: Just a way to get to the element to add the pastie. If there's a way to get the current message li to append to, then that would be better.

-----Original Message-----

PR #55, opened 2011-10-30
Update to make images and videos collapsible

ruddles:


Chat/ContentProviders/PastieContentProvider.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Text.RegularExpressions;
+using System.Text.RegularExpressions;
```
davidfowl: Sort usings. (for all files)

PR #61, opened 2011-11-08
Add TAB and Shift-TAB support to autocomplete

kiliman:
- Note: values are only sorted if a prefix is supplied
- Otherwise it will cycle in the order the nicknames are listed

### DIFF
--- a/Chat/Scripts/jquery.autotabcomplete.js
+++ b/Chat/Scripts/jquery.autotabcomplete.js
@@ -35,7 +35,7 @@
                     _text = $(this).val();
 
                     // find prefix (starts with @)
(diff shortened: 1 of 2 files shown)


Chat/Scripts/jquery.autotabcomplete.js
```diff
@@ -14,61 +14,84 @@
...
-        var _text = '';
-        var _prefix = '';
-        var _index = 0;
+        var _inAutoComplete;
+        var _text;
+        var _prefix;
+        var _index;
+
+        var KEY = {
+		    TAB: 9,
```
davidfowl, line 23: formatting and why the all caps (this ain't c++ :))

PR #68, opened 2011-11-12
Added list command

ruddles:
Added /list command to that lists the users in a room

### DIFF
--- a/Chat/Chat.cs
+++ b/Chat/Chat.cs
@@ -474,7 +474,7 @@ namespace SignalR.Samples.Hubs.Chat
             }
 
             var names = room.Users.Select(s => s.Name);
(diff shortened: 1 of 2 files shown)


Chat/Chat.cs
```diff
@@ -454,6 +459,24 @@ namespace SignalR.Samples.Hubs.Chat
...
+
+            var room = _db.Rooms.FirstOrDefault(r => r.Name.Equals(parts[1], StringComparison.OrdinalIgnoreCase));
+
+            if (room == null)
+            {
+                throw new InvalidOperationException("No room with that name!");
+            }
+
+            var names = room.Users.Select(s => s.Name);
+            Caller.showUsersInRoom(room.Name, names.ToArray());
```
davidfowl: Why the ToArray call?

PR #70, opened 2011-11-14
Added a chat history

Buildstarted:
Added a chat history to the text input that allows you to press up/down to cycle through previous messages


Chat.1.1.dotCover
```diff
@@ -0,0 +1,11 @@
+<Configuration>
+  <CoverageFilters>
+    <IncludeFilters>
+      <Filter ModuleMask="*" ClassMask="*" FunctionMask="*" />
+    </IncludeFilters>
+    <ExcludeFilters />
+  </CoverageFilters>
+  <SnapshotDialog>
+    <InitialDirectory>H:\Dev\SignalR-Chat</InitialDirectory>
```
davidfowl: Uh what is H:\? I don't have that drive. :)

Buildstarted: I thought I removed all the resharper/jetbrains things. The lack of exceptions in .gitignore and my affinity for "git add ." when i don't need to always use it - I need to delete that file.

Chat/Chat.js
```diff
@@ -486,6 +499,17 @@ $(function () {
             }
         });
     });
+
+    //Chat history setup
+    var chatHistory = new Array();
```
davidfowl: I'm not sure that new Array()? Change it to:

```
chatHistory = [];
```

Chat/Chat.js
```diff
@@ -389,12 +390,24 @@ $(function () {
...
 
+        // cycle through the history 
+        var key = (e.keyCode ? e.keyCode : e.which);
+        if (key == 38) {
+            historyLocation -= 1;
+            if (historyLocation < 0) historyLocation = 0;
+            $(this).val(chatHistory[historyLocation]);
+        } else if (key == 40) {
+            historyLocation += 1;
+            if (historyLocation > history.length) historyLocation = chatHistory.length;
```
davidfowl: No self closing ifs

Chat/Chat.js
```diff
@@ -389,12 +390,24 @@ $(function () {
             chat.typing(true);
         }
 
+        // cycle through the history 
+        var key = (e.keyCode ? e.keyCode : e.which);
```
davidfowl, line 395: jQuery normalizes this event so you should be able to use e.keyCode regardless (I believe)

Chat/Chat.js
```diff
@@ -389,12 +390,24 @@ $(function () {
             chat.typing(true);
         }
 
+        // cycle through the history 
+        var key = (e.keyCode ? e.keyCode : e.which);
+        if (key == 38) {
+            historyLocation -= 1;
+            if (historyLocation < 0) historyLocation = 0;
+            $(this).val(chatHistory[historyLocation]);
+        } else if (key == 40) {
```
davidfowl: Use `===` instead of `==`.

Chat/Chat.js
Buildstarted, line 395: I had check e.which (which was what the jquery docs used) but I found it doesn't work in FF 4.

Chat/Chat.js
```diff
@@ -389,12 +390,28 @@ $(function () {
             chat.typing(true);
         }
 
+        // cycle through the history 
+        var key = (e.keyCode ? e.keyCode : e.which);
+        if (key === 38) {
+            historyLocation -= 1;
+            if (historyLocation < 0) {
```
davidfowl, line 398: The formatting below here looks really messed up.

Chat/Chat.js
```diff
@@ -389,12 +390,28 @@ $(function () {
             chat.typing(true);
         }
 
+        // cycle through the history 
+        var key = (e.keyCode ? e.keyCode : e.which);
+        if (key === 38) {
```
davidfowl: Don't use magic numbers. Set some constants like Keys.Down, Keys.Up:

```
var Keys = { Up : 38 }
```

Something like that.

Chat/Chat.js
Buildstarted, line 398: I just realized that notepad++ used tabs when vs used spaces. sorry - will fix

Chat/Chat.js
davidfowl, line 395: So you need this logic? If you do change it to:

var key = e.keyCode || e.which;

I'm surprised you do though..

Chat/Chat.js
```diff
@@ -389,12 +391,28 @@ $(function () {
...
+        if (key === Keys.Up) {
+            historyLocation -= 1;
+            if (historyLocation < 0) {
+                historyLocation = 0;
+            }
+            $(this).val(chatHistory[historyLocation]);
+        } else if (key === Keys.Down) {
+            historyLocation += 1;
+            if (historyLocation > history.length) {
+                historyLocation = chatHistory.length;
```
davidfowl, line 405: Should this be historyLocation - 1?

The above logic is clear but I'd write:

```
historyLocation = Math.min(historyLocation + 1, chatHistory.length - 1);
```

But not a big deal, just a style thing.

Chat/Chat.js
Buildstarted, line 395: You were right, the docs stated to use e.which but e.keyCode performs as expected.

Chat/Chat.js
Buildstarted, line 405: The reason it's HistoryLocation = chatHistory.length is so that when you get to the bottom it clears out what you were typing so that the textbox is empty. probably a bad use of that.

davidfowl, line 405: I figured but maybe you want to make that more explicit instead of intentionally going out of bounds.

PR #83, opened 2011-11-16
Added /who command

ruddles:
Added /who command:

/who - lists all users
/who [name] - lists rooms the user is in.

### DIFF
--- a/Chat/Chat.cs
+++ b/Chat/Chat.cs
@@ -465,7 +465,8 @@ namespace SignalR.Samples.Hubs.Chat
 
             if (users.Count() == 1)
             {
-                Caller.showUsersRoomList(users.First().Name, users.First().Rooms.Select(r => r.Name));
+                var user = users.First();
+                Caller.showUsersRoomList(user.Name, user.Rooms.Select(r => r.Name));
             }
             else
             {
--- a/Chat/Chat.cs
+++ b/Chat/Chat.cs
@@ -382,6 +382,11 @@ namespace SignalR.Samples.Hubs.Chat
                     HandleList(parts);
                     return true;
                 }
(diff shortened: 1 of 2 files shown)


Chat/Chat.cs
```diff
@@ -438,6 +443,36 @@ namespace SignalR.Samples.Hubs.Chat
...
+
+            var exactUserMatch = _db.Users.FirstOrDefault(s => s.Name.Equals(name, StringComparison.OrdinalIgnoreCase));
+
+            if (exactUserMatch != null)
+            {
+                Caller.showUsersRoomList(exactUserMatch.Name, exactUserMatch.Rooms.Select(r => r.Name));
+                return;
+            }
+
+            var users = _db.Users.Where(s => s.Name.IndexOf(name, StringComparison.OrdinalIgnoreCase) != -1);
```
davidfowl, line 464: Why IndexOf and note Equals? Ah it's a search

Chat/Chat.cs
```diff
@@ -438,6 +443,36 @@ namespace SignalR.Samples.Hubs.Chat
...
+
+            if (users.Count() == 1)
+            {
+                Caller.showUsersRoomList(users.First().Name, users.First().Rooms.Select(r => r.Name));
```
davidfowl: Change this to:

```
var user = users.First();
Caller.showUsersRoomList(user.Name, user.Rooms.Select(r => r.Name));
```

Chat/Chat.cs
ruddles, line 464: Aye, allows partial, case-insensitive, matches

Chat/Chat.cs
ruddles: Done

Issue #97, opened 2011-11-28
/kick command doesn't accept @nickname


kiliman:
When you try to /kick someone and use the @nickname form, it says "Couldn't find any user named '@nickname'"

Should allow @nicknames.

kiliman:
Sorry for the duplicates commits... pull requests are including spurious commits (merges, etc.).

I created an upstream branch that I'm now basing my changes on, so hopefully you'll get clean pull requests.

PR #98, opened 2011-11-28
Fix issue #97 - /kick command doesn't accept @nickname

kiliman:


JabbR/Hubs/Chat.cs
```diff
@@ -509,7 +509,7 @@ namespace JabbR
                 throw new InvalidOperationException("You're the only person in here...");
             }
 
-            string targetUserName = parts[1];
+            string targetUserName = NormalizeUserName(parts[1]);
```
davidfowl: Why not move this to EnsureUser?

PR #116, opened 2011-11-30
Adding unit tests for command manager

daanl:
nick
join
addowner

### DIFF
--- a/JabbR.Test/CommandManagerFacts.cs
+++ b/JabbR.Test/CommandManagerFacts.cs
@@ -146,8 +146,7 @@ namespace JabbR.Test
                 var user = new ChatUser
                 {
                     Name = "dfowler",
-                    Id = "1",
-                    HashedPassword = "password".ToSha256()
+                    Id = "1"
                 };
                  
                 repository.Add(user);
@@ -322,8 +321,7 @@ namespace JabbR.Test
                 var user = new ChatUser
                 {
                     Name = "dfowler",
-                    Id = "1",
-                    HashedPassword = "password".ToSha256()
+                    Id = "1"
                 };
                 repository.Add(user);
                 var room = new ChatRoom
@@ -353,8 +351,7 @@ namespace JabbR.Test
                 var user = new ChatUser
                 {
                     Name = "dfowler",
-                    Id = "1",
-                    HashedPassword = "password".ToSha256()
+                    Id = "1"
                 };
                 repository.Add(user);
                 var room = new ChatRoom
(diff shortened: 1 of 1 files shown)


JabbR.Test/CommandManagerFacts.cs
```diff
@@ -287,6 +314,155 @@ namespace JabbR.Test
...
+                bool result = commandManager.TryHandleCommand("/join room");
+
+                Assert.True(result);
+                notificationService.Verify(m => m.JoinRoom(user, room), Times.Once());
+            }
+
+            [Fact]
+            public void ThrowIfRoomIsEmpty()
+            {
+                var repository = new InMemoryRepository();
```
davidfowl, line 350: This can be reduced to VerifyThrows<InvalidOperationException>("/join");

JabbR.Test/CommandManagerFacts.cs
```diff
@@ -287,6 +314,155 @@ namespace JabbR.Test
...
+
+                Assert.Throws<InvalidOperationException>(() => commandManager.TryHandleCommand("/join"));
+            }
+        }
+
+        public class AddOwnerCommand
+        {
+            [Fact]
+            public void MissingUserNameThrows()
+            {
```
davidfowl, line 382: Same as above.

JabbR.Test/CommandManagerFacts.cs
```diff
@@ -287,6 +314,155 @@ namespace JabbR.Test
...
+                                                        repository,
+                                                        notificationService.Object);
+
+                Assert.Throws<InvalidOperationException>(() => commandManager.TryHandleCommand("/addowner "));
+            }
+
+            [Fact]
+            public void MissingRoomNameThrows()
+            {
+                var repository = new InMemoryRepository();
```
davidfowl, line 405: And here

JabbR.Test/CommandManagerFacts.cs
```diff
@@ -287,6 +314,155 @@ namespace JabbR.Test
...
+                                                        "1",
+                                                        null,
+                                                        service,
+                                                        repository,
+                                                        notificationService.Object);
+
+                var result = commandManager.TryHandleCommand("/addowner dfowler2 test");
+
+                Assert.True(result);
+                notificationService.Verify(m => m.OnOwnerAdded(targetUser, room), Times.Once());
```
davidfowl, line 460: You didn't very that the objects were updated (i.e that dfowler2 is has the "test" room in the owner collection and vice versa "test" has dfowler2 in the owners collection).

JabbR.Test/CommandManagerFacts.cs
daanl, line 350: VerifyThrows doesn't has an user, so it won't reach the code. 

ChatUser user = _repository.VerifyUserId(_userId); will throw an exception if there is no user. 

With exception: "You don't have a name. Pick a name using '/nick nickname'."

I could create an overload for VerifyThrows with parameters command, user and one other overload with parameters: command, user and room. So its re-usable at other places.

PR #123, opened 2011-12-01
added notice to title bar when unread messages to specific user

Buildstarted:
Instead of seeing "(5) chatroom" it will show "*(5) chatroom" if that room has a highlighted message to the logged in user.

### DIFF
--- a/JabbR/Chat.js
+++ b/JabbR/Chat.js
@@ -11,6 +11,7 @@
         historyLocation = 0,
         originalTitle = document.title,
         unread = 0,
+        unreadNotice = false,
         focus = true,
         typingTimeoutId = null;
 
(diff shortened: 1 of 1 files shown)


JabbR/Chat.js
```diff
@@ -98,12 +99,18 @@
...
         else {
-            document.title = '(' + unread + ') ' + originalTitle;
+            document.title =
+                (unreadNotice ? '*' : '')
+                + '(' + unread + ') ' + originalTitle;
         }
     }
 
-    function updateUnread(room) {
+    function updateUnread(room, viewModel) {
```
davidfowl, line 108: Can't you decide if  the message is for the current client outside of this method and pass something more focused like:

```
function updateUnread(room, isMention) {
}
```

And before you call it, you can do:

```
var isMention = viewModel.highlight === 'highlight'; 
```

I'd rather change the view model to have a property called isMention since it already knows that way the logic is cleaner:

```
if(focus === false) {
    hasUnreadMentions =  isMention;
}
```

Something along those lines

PR #148, opened 2011-12-02
added hashtag parsing to messages to create links to rooms

Buildstarted:
automatically parses #hashtags and converts them to links to rooms. It does not convert any rooms that don't exist.


davidfowl:
Add some unit tests

PR #149, opened 2011-12-02
INotificationService.ListUsers() ordering alphabetically

Buildstarted:


davidfowl:
Change looks incomplete. What happens when users get added/removed or change names?

PR #153, opened 2011-12-02
Add slideshare content provider

sethwebster:
Isolated changes; switched out WebClient for HttpWebRequest -- added some comments on why it's implemented as a separate request.

Basically, slidehsare's embed requires a unique id which comes from the oEmbedApi call.

### DIFF
--- a/JabbR/ContentProviders/SlideshareContentProvider.cs
+++ b/JabbR/ContentProviders/SlideshareContentProvider.cs
@@ -14,16 +14,6 @@ namespace JabbR.ContentProviders
     /// </summary>
     public class SlideShareContentProvider : IContentProvider
     {
-
-        private static readonly string slideScript = string.Format(
-            "<div style='width:425px' id='__ss_10388027'> " +
-            "<strong style='display:block;margin:12px 0 4px'>" +
-            "<a href='{0}' target='_blank'></a></strong> " +
-           " <iframe src='http://www.slideshare.net/slideshow/embed_code/10388027' width='425' height='355' frameborder='0' marginwidth='0' marginheight='0' scrolling='no'></iframe> <div style='padding:5px 0 12px'> View more <a href='http://www.slideshare.net/' target='_blank'>presentations</a> from <a href='http://www.slideshare.net/verticalmeasures' target='_blank'>Vertical Measures</a> </div> </div>",
-           "test"
-
-            );
-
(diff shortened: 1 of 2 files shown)


JabbR/ContentProviders/SlideshareContentProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+              || response.ResponseUri.AbsoluteUri.StartsWith("http://www.slideshare.net/", StringComparison.OrdinalIgnoreCase))
+            {
+
+                try
+                {
+                    // We have to make a call to the SlideShare api because
+                    // their embed code request the unique ID of the slide deck
+                    // where we will only have the url -- this call gets the json information
+                    // on the slide deck and that package happens to already contain the embed code (.html)
+                    HttpWebRequest webRequest = (HttpWebRequest)HttpWebRequest.Create(
```
davidfowl: var

JabbR/ContentProviders/SlideshareContentProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+                            string.Format(oEmbedUrl, response.ResponseUri.AbsoluteUri));
```
davidfowl: Uppercase S

JabbR/ContentProviders/SlideshareContentProvider.cs
```diff
@@ -0,0 +1,51 @@
...
+                    using (var reader = new StreamReader(webResponse.GetResponseStream()))
+                    {
+                        dynamic slideShareData = JsonConvert.DeserializeObject(reader.ReadToEnd()
+                                   );
+                        return slideShareData.html;
+                    }
+                }
+                catch (Exception ex)
+                {
+                    throw ex;
```
davidfowl: Why are you rethrowing at all? 
What happens to chat in this case? 
When does this happen? 
Why not just return null?

sethwebster: What's the right thing to do here?  Is there a mechanism to inform chat that the request failed? Something more informative than nothing?

davidfowl: You should test this code path. I think we just write the out error out to the trace but you should verify your expectations. Also if you rethrow why do you catch at all?

sethwebster:
Updated with suggestions.

PR #158, opened 2011-12-03
Add vimeo content provider

sethwebster:


JabbR/ContentProviders/VimeoContentProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace JabbR.ContentProviders
+{
+    public class VimeoContentProvider : EmbedContentProvider
+    {
+        protected override IEnumerable<object> ExtractParameters(Uri responseUri)
+        {
+            if (responseUri.Segments.Length > 1)
+            {
+                string videoId = responseUri.Segments[1];
```
davidfowl: Why not use a regex like the rest of these?

JabbR/ContentProviders/VimeoContentProvider.cs
```diff
@@ -0,0 +1,47 @@
...
+                    yield return videoId;
+                }
+            }
+            yield return "";
+        }
+
+
+        public override System.Collections.Generic.IEnumerable<string> Domains
+        {
+            get { return new String[] { "http://vimeo.com", "http://www.vimeo.com" }; }
```
davidfowl: Use yield return instead of an array.

JabbR/ContentProviders/VimeoContentProvider.cs
```diff
@@ -0,0 +1,47 @@
+﻿
+
+using System;
+using System.IO;
+using System.Net;
+using Newtonsoft.Json;
+using System.Collections.Generic;
```
davidfowl: Remove and sort usings.

JabbR/ContentProviders/VimeoContentProvider.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System.Text.RegularExpressions;
+
+namespace JabbR.ContentProviders
+{
+    public class VimeoContentProvider : EmbedContentProvider
+    {
+        private static readonly Regex _vimeoIdRegex = new Regex(@"(\d+)");
+
+        public override System.Text.RegularExpressions.Regex MediaUrlRegex
```
davidfowl: Why are full type names here. Add the namespace import.

PR #162, opened 2011-12-03
parsing hashtag links in private messages

Buildstarted:
added support for hashtag parsing in private messages


JabbR/Commands/CommandManager.cs
```diff
@@ -347,6 +347,9 @@ private void HandleMsg(ChatUser user, string[] parts)
                 throw new InvalidOperationException(String.Format("What did you want to say to '{0}'.", toUser.Name));
             }
 
+            TextTransform transform = new TextTransform(_repository);
```
davidfowl: var

PR #163, opened 2011-12-03
used same regex as room validator for hashtags

Buildstarted:
should be refactored later so that one change will affect them all but it's probably not necessary since there's no reason to change now

### DIFF
--- a/JabbR/Infrastructure/TextTransform.cs
+++ b/JabbR/Infrastructure/TextTransform.cs
@@ -11,7 +11,8 @@ namespace JabbR.Infrastructure {
     public class TextTransform 
     {
         private readonly IJabbrRepository _repository;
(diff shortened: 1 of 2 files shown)


JabbR/Infrastructure/TextTransform.cs
```diff
@@ -11,7 +11,8 @@ namespace JabbR.Infrastructure {
     public class TextTransform 
     {
         private readonly IJabbrRepository _repository;
-        public const string HashTagPattern = @"(?:(?<=\s)|^)#(\w*[A-Za-z_]+\w*)";
+        public const string HashTagPattern = @"(?:(?<=\s)|^)#([A-Za-z0-9-_.]{1,30}\w*)";
+        //public const string HashTagPattern = @"(?:(?<=\s)|^)#(\w*[A-Za-z_-]+\w*)";
```
davidfowl: Can you get rid of this commented out variable?

PR #165, opened 2011-12-03
Add google presentation content provider

sethwebster:


JabbR/ContentProviders/GoogleDocsFormProvider.cs
```diff
@@ -19,8 +19,8 @@ public override IEnumerable<string> Domains
         {
             get
             {
-                yield return "https://docs.google.com";
```
davidfowl, line 22: Don't get this change.

sethwebster, line 22: the URL for the forms includes with spreadsheet and in adding the presentations content provider, we need to be able to differentiate between spreadsheets and presentations

davidfowl, line 22: ok

Issue #169, opened 2011-12-03
GooglePresentationsContentProvider doesn't work with the obvious url the author sees when creating


sethwebster:
The URL that will work with the GooglePresentationsContentProvider is the one that is given when a "Publish to web" operation is performed on the Google Presentation.  If the author uses the obvious URL seen while editing the presentation as an owner, a login is required.

PR #170, opened 2011-12-03
Fix for issue #169 -- Added a special case where login redirect is found the id is pulled from the url in a different spot

sethwebster:


JabbR/ContentProviders/GoogleDocsPresentationsContentProvider.cs
```diff
@@ -1,12 +1,15 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Net;
 using System.Web;
+using System.Text.RegularExpressions;
```
davidfowl, line 5: Sort usings.

Issue #171, opened 2011-12-04
ContentProviders should return a model v.s. a string 


sethwebster:
This will allow us to give a richer experience to the inline content... Titles and management of the content for instance

PR #172, opened 2011-12-04
Added ContentProviderResultModel as the return type from providers (for issue #171)

sethwebster:
All providers tested and working at http://seth-jabbr.apphb.com


sethwebster:
Moved ContentProviderResultModel into the JabbR.ContentProviders namespace/folder

PR #174, opened 2011-12-04
Converted ContentProviderResultModel into struct and gave a convention for returning an empty result in ContentProviderResultModel.Empty

sethwebster:


JabbR/Hubs/Chat.cs
```diff
@@ -287,7 +287,8 @@ private void ProcessUrls(IEnumerable<string> links, ChatRoom room, ChatMessage c
                         continue;
                     }
 
-                    if (task.Result == null || String.IsNullOrEmpty(task.Result.Content))
+                    
+                    if (task.Result == ContentProviderResultModel.Empty)
```
davidfowl: Just check the content. It can go back to:

`if(String.IsNullOrEmpty(task.Result.Conent))`

JabbR/ContentProviders/ContentProviderResultModel.cs
```diff
@@ -1,9 +1,34 @@
...
+            return toCompare.Content == this.Content &&
+                toCompare.Title == this.Title;
+        }
+
+        public override int GetHashCode()
+        {
+            return base.GetHashCode();
+        }
+
+        public static bool operator ==(ContentProviderResultModel left, ContentProviderResultModel right)
```
davidfowl: The operator overloading is overkill.

JabbR/ContentProviders/ContentProviderResultModel.cs
```diff
@@ -1,9 +1,34 @@
...
+            if (!(obj is ContentProviderResultModel))
+                return false;
+            var toCompare = (ContentProviderResultModel)obj;
+            return toCompare.Content == this.Content &&
+                toCompare.Title == this.Title;
+        }
+
+        public override int GetHashCode()
+        {
+            return base.GetHashCode();
```
davidfowl: What?

JabbR/ContentProviders/ContentProviderResultModel.cs
```diff
@@ -1,9 +1,34 @@
...
 namespace JabbR.ContentProviders
 {
-    public class ContentProviderResultModel
+    public struct ContentProviderResultModel
     {
         public string Title { get; set; }
         public string Content { get; set; }
+        public static ContentProviderResultModel Empty = new ContentProviderResultModel();
+
+        public override bool Equals(object obj)
```
davidfowl: Why override all of these?

sethwebster:
I figured the code reads nicer with .Empty -- makes sense when read.

davidfowl:
Maybe leaving it as null was simpler. I think thats all over kill. The whole goal was to prevent having to check for null, nothing else.

PR #175, opened 2011-12-04
Add joinme content provider

sethwebster:
Content Provider that allows Join.me screen shares to be embedded in chats...

### DIFF
--- a/JabbR/ContentProviders/JoinMeContentProvider.cs
+++ b/JabbR/ContentProviders/JoinMeContentProvider.cs
@@ -1,0 +1,38 @@
+﻿using System;
+using System.Linq;
+using System.Net;
+using System.Text.RegularExpressions;
+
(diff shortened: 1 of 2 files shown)


JabbR/ContentProviders/JoinMeContentProvider.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Linq;
+using System.Net;
+using System.Text.RegularExpressions;
+
+namespace JabbR.ContentProviders
+{
+    public class JoinMeContentProvider : CollapsibleContentProvider
+    {
+        private static readonly Regex _joinMeIdRegex = new Regex(@"(\d+)");
+        // 0 Lat, 1 Long, 2 Address, 3 Content Format for Info
```
davidfowl, line 11: This doesn't belong here. Too much copy paste ;)

PR #180, opened 2011-12-05
Tests for command manager

daanl:


JabbR.Test/CommandManagerFacts.cs
```diff
@@ -6,6 +6,8 @@
 using JabbR.Services;
 using Moq;
 using Xunit;
+using System.Collections.Generic;
+using System.Threading;
```
davidfowl: Why?

JabbR/Commands/CommandManager.cs
```diff
@@ -30,17 +30,14 @@ public class CommandManager
             _notificationService = notificationService;
         }
 
-        public bool TryHandleCommand(string command)
+        public bool TryHandleCommand(string commandName, string[] parts)
```
davidfowl, line 33: Why is this public?

JabbR.Test/CommandManagerFacts.cs
daanl: using System.Collections.Generic; 
Is used in line 1282 (var userList = new List<String>();)

using System.Threading; is now removed (updated pull request)

JabbR/Commands/CommandManager.cs
daanl, line 33: Otherwise there are some untestable lines for example:

string userName = parts[1];
            if (String.IsNullOrWhiteSpace(userName))
            {
                throw new InvalidOperationException("No nick specified!");
            }

These lines in HandleNick can not be touched due .net Split command it doesn't split two spaces.

JabbR.Test/CommandManagerFacts.cs
```diff
@@ -6,6 +6,7 @@
 using JabbR.Services;
 using Moq;
 using Xunit;
+using System.Collections.Generic;
```
davidfowl: Sort usings.

daanl: Done

PR #181, opened 2011-12-06
Added UstreamContentProvider

sethwebster:
I know, it's not a Sprocket.

### DIFF
--- a/JabbR/ContentProviders/UStreamContentProvider.cs
+++ b/JabbR/ContentProviders/UStreamContentProvider.cs
@@ -1,0 +1,67 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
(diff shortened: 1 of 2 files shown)


JabbR/ContentProviders/UStreamContentProvider.cs
```diff
@@ -0,0 +1,67 @@
...
+                {
+                    dynamic ustreamAssetData = JsonConvert.DeserializeObject(reader.ReadToEnd());
+                    return ustreamAssetData;
+                }
+            }
+        }
+
+
+
+        protected IEnumerable<string> ExtractParameters(Uri responseUri)
```
davidfowl, line 50: This method is duplicated way too many times. Also, maybe return IList instead so you don't have to use ElementAt

PR #202, opened 2011-12-08
Chrome popup notifications

ajryan:
Hi! Thanks for creating JabbR!

First off, I am a major open source and git newbie, sorry if this is a mess. I implemented Chrome popup notifications that appear only when the JabbR window is not focused.

Notes:
1. Works only on Chrome
2. Click "Enable notifications" to grant notification permission to the domain where JabbR is hosted. The "enable" link is in a silly place.
3. A popup will appear when a new message is received while !ui.focus. It will stay visible until:
   
   a. It is explicitly closed OR
   b. Another "background" message is received, in which case a new popup will replace the previous one.

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -9,12 +9,14 @@
         $tabs = null,
         $submitButton = null,
         $newMessage = null,
+        $enableNotifications = null,
         templates = null,
         app = null,
         focus = true,
         commands = [],
         Keys = { Up: 38, Down: 40, Esc: 27 },
-        scrollTopThreshold = 5;
+        scrollTopThreshold = 5,
+        chromePopup = null;
 
     function getRoomId(roomName) {
         return escape(roomName.toLowerCase()).replace(/[^a-z0-9]/, '_');
@@ -315,6 +317,7 @@
             $tabs = $('#tabs');
             $submitButton = $('#send-message');
             $newMessage = $('#new-message');
+            $enableNotifications = $('#enable-notifications');
             focus = true;
             templates = {
                 user: $('#new-user-template'),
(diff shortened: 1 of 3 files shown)


JabbR/Chat.ui.js
```diff
@@ -399,6 +402,12 @@
                 return false;
             });
 
+            $enableNotifications.click(function () {
+                if (window.webkitNotifications) {
+                    window.webkitNotifications.requestPermission(function () { alert('Notification permissions granted.'); });
```
davidfowl: Alert? I rather do nothing for now.

ajryan: Indeed, nothing is better. In the end, dynamic [enable notifications] that transitions to green [notifications enabled] or red [failed].

JabbR/Chat.ui.js
```diff
@@ -706,6 +715,22 @@
                   .find('.right').remove(); // remove timestamp on date indicator
             }
 
+            // when we are not focused, attempt chrome notifications
+            if (!ui.focus) {
```
davidfowl: I'd rather have this in updateUnread but you want to get the actual message. Can you post of what it looks like? 

Also can you refactor this into a private function?

`function notifyMessage(message) {`
`}`

JabbR/index.htm
```diff
@@ -90,6 +90,9 @@
             <h1>Jabb</h1>
             <img src="Content/images/logo32.png" id="logo" />
             <div>Powered by <a href="https://github.com/SignalR/SignalR" target="_blank">SignalR</a></div>
+            <div>
+                <a id="enable-notifications"><span class="content">Enable notifications</span></a>
```
davidfowl: I'd like to see what this looks like as well.

PR #207, opened 2011-12-09
Adding Sound Notifications

osbornm:
Adding a notification sound when you get a pm & when you get a mention in an inactive room or when the window does not have focus.


davidfowl:
Can you redo this pull request and fix the line endings? Alternatively you can try to do a git commit --amend.

osbornm:
You should pull this because I am the one that actually fixed your line endings to the right ones :)

PR #212, opened 2011-12-11
Not sure why but the Twitter Regex wasn't working for all links; saw some broken in chat so I changed regex.

sethwebster:


JabbR/ContentProviders/TweetContentProvider.cs
```diff
@@ -17,7 +17,7 @@ public class TweetContentProvider : CollapsibleContentProvider
         /// <summary>
         /// Regex for parsing the tweet ID out of the link.
         /// </summary>
-        private static readonly Regex tweetRegex = new Regex(@".*https?://twitter.com/(?:#!/)*[^/]+/status/(\d+).*");
+        private static readonly Regex _tweetRegex = new Regex(@".*/statuses/(\d+)");
```
sethwebster, line 20: Also added an underscore since @dantup forgot his :)

davidfowl, line 20: This looks drastically simplified. Works well?

PR #217, opened 2011-12-11
Fixed missing reference to pinned windows and re-added merged out #page element

sethwebster:
Looks like the page element and script reference were merged out


davidfowl:
What happened to the formatting? Also you're using spaces right no tabs?

sethwebster:
I got it this way this time -- you want me to reformat it?

PR #218, opened 2011-12-11
More js refactoring

redsquare:
pass in document object into module construct to allow for better munging,
also cache and use jq document object rather than create multiple instances


redsquare:
ignore this - it was done late and includes some of your (formatting) fixes already

davidfowl:
Tmake sure your formatting matches mine. This chanes causes merge conflicts

Issue #231, opened 2011-12-14
Need a way to limit room number per user


jharjadi:
To prevent user to spam by creating too many room, we need to have a column in user table something like "room-allowance" and by default can be set like 5 or 10.

so erverytime command /create triggered need to check this room-allowance number against existing room.

PR #240, opened 2011-12-15
Add room allowance for user max 5, update addRoom service to check room ...

jharjadi:
...allowance. [#231]

Add max-weight for image in the collapsible-box (prevent displaying super big image)


davidfowl:
So once I hit the limit I can never create another room?

davidfowl:
Before accepting this we'd need to have the ability to close rooms.

jharjadi:
Will need admin function/command to be able to set this RoomAllowance.

davidfowl:
That's orthogonal to this but yes, we'll need that as well. Anyway this feature is premature, we need other features to make this work well.

PR #241, opened 2011-12-15
Add github issues content provider

sethwebster:


JabbR/Scripts/Markdown.Converter.js
```diff
@@ -0,0 +1,1332 @@
```
davidfowl, line 0: What the license?

sethwebster, line 0: MIT

sethwebster, line 0: http://code.google.com/p/pagedown/source/browse/LICENSE.txt

PR #256, opened 2011-12-17
PM everywhere is working - fixes #72

### LINKED ISSUE #72: MSG command should repeat to each joined room.
/msg should be sent to all joined tabs to avoid missing messages.

sethwebster:


JabbR/Chat.ui.js
```diff
@@ -221,6 +221,16 @@
                         $('.messages.current'));
     }
 
+    function getAllRoomElements() {
+        var rooms = [];
+        $("ul#tabs > li.room")  .each(function () {
```
davidfowl, line 226: Isnt there a getRoomElements function? Why do new Room()?

sethwebster, line 226: Not one that gets the list of current rooms (that I found)... getRoomElements gets the elements of a single room.  As for new Room(), I was following the existing pattern and that would be what I expected to receive from such a call.

davidfowl, line 226: There's one that gets it by name so no need to manually use selectors.

sethwebster, line 226: You mean within the new Room(<in here>)

PR #262, opened 2011-12-20
Issue#237

PureKrome:
Here we go :)

This Pull Requests is for the following added functionality:
- /Note <optional text>
- /Afk <optional text>

EDIT: I've also done a merge with upstream/master .. merge 2x conflicts .. committed the Merge .. and now Re-Pull Requested.


JabbR/Chat.js
```diff
@@ -393,6 +397,20 @@
...
+    chat.noteChanged = function (isAfk, isCleared) {
+        ui.addMessage(isAfk ? 'You have gone Afk' : 'Your note has been ' + (isCleared ? 'cleared' : 'set'), 'notification', this.activeRoom);
+    };
+
+    // Make sure all the people in all the rooms know that a user has changed their note.
+    chat.changeNote = function (user, room) {
+        ui.changeNote(user, room);
+
+        if (!isSelf(user)) {
+            ui.addMessage(user.Name + " has " + (user.IsAfk ? "gone Afk" : (user.Note == null ? "cleared" : "set") + " their note"), 'notification', room);
```
davidfowl: don't use double quotes, everywhere else uses single quotes.

Also clean this line up, way too much expressions in here, makes it unreadable

JabbR/Chat.ui.js
```diff
@@ -1060,9 +1066,25 @@
...
+        changeNote: function (user, roomName) {
+            var room = getRoomElements(roomName),
+                $user = room.getUser(user.Name);
+
+            toggleNote(user, $user);
         }
     };
 
+    function toggleNote(user, $user) {
+        var element = $user.find('.note');
```
davidfowl: We use $element for references jQuery elements. In this case I'd call the variable $note.

JabbR/Commands/CommandManager.cs
```diff
@@ -766,5 +779,25 @@ private void HandleNudge(ChatRoom room, ChatUser user, string[] parts)
                 throw new InvalidOperationException(String.Format("Room can only be nudged once every {0} seconds", betweenNudges.TotalSeconds));
             }
         }
+
+        private void HandleNote(ChatUser user, string[] parts)
+        {
+            // NOTE (see what i did there!): "Parts" with only 1 item means we're clearing the note.
```
davidfowl: What is this comment?

Can you make this code clearer.

JabbR/Hubs/Chat.cs
```diff
@@ -682,6 +684,25 @@ void INotificationService.OnUserNameChanged(ChatUser user, string oldUserName, s
             }
         }
 
+        void INotificationService.ChangeNote(ChatUser user)
+        {
+            // Update the calling client
+            foreach (var client in user.ConnectedClients)
+            {
+                Clients[client.Id].noteChanged(user.IsAfk, user.Note == null);
```
davidfowl: calculate `user.Note == null` outside of the loop and give it a proper variable name.

JabbR/Models/ChatUser.cs
```diff
@@ -18,6 +18,13 @@ public class ChatUser
         public DateTime LastActivity { get; set; }
         public DateTime? LastNudged { get; set; }
         public int Status { get; set; }
+        public string Note { get; set; }
+
+        [NotMapped]
+        public bool IsAfk
+        {
+            get { return Note != null && Note.StartsWith("Afk"); }
```
davidfowl: Ignore case for the starts with? Maybe use a constant?

JabbR/ViewModels/UserViewModel.cs
```diff
@@ -10,11 +10,16 @@ public UserViewModel(ChatUser user)
...
+            Note = user.Note;
+            IsAfk = user.IsAfk;
         }
 
         public string Name { get; set; }
         public string Hash { get; set; }
         public bool Active { get; set; }
         public string Status { get; set; }
+        public string Note { get; set; }
+        public string NoteCss { get { return Note == null ? null : IsAfk ? "afk" : "message"; } }
```
davidfowl: Split this into multiple lines. Not sure about putting the css class here, can't this be done in JavaScript?

PureKrome: It can.. but IMO the View (which i see the JS, is) should not have any logic in it.

JabbR/JabbR.csproj
```diff
@@ -45,7 +45,7 @@
     <Reference Include="EntityFramework">
       <HintPath>..\packages\EntityFramework.4.2.0.0\lib\net40\EntityFramework.dll</HintPath>
     </Reference>
-    <Reference Include="EntityFramework.Migrations">
+    <Reference Include="EntityFramework.Migrations, Version=0.7.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089, processorArchitecture=MSIL">
```
davidfowl: Did you upgrade?

JabbR/ViewModels/UserViewModel.cs
davidfowl: IMO it can, there's different layers of view models. Take a look at all the other crap we do in javascript.

JabbR/Chat.js
PureKrome: Is this considered, cleaned up / easier to read?

``` javascript
// Called when you have added or cleared a note
chat.noteChanged = function (isAfk, isCleared) {
    var afkMessage = 'You have gone Afk';
    var noteMessage = 'Your note has been ' + isCleared ? 'cleared' : 'set';
    ui.addMessage(isAfk ? afkMessage: noteMessage, 'notification', this.activeRoom);
};
```

davidfowl: Yup.

JabbR/Chat.ui.js
PureKrome: Awesome - done.

JabbR/Chat.ui.js
```diff
@@ -1060,9 +1066,25 @@
...
+
+            toggleNote(user, $user);
         }
     };
 
+    function toggleNote(user, $user) {
+        var $note = $user.find('.note');
+        $note.removeClass('afk message');
+        $note.removeAttr('title');
+        if (user.Note != null) {
```
davidfowl: `!==`

JabbR/Commands/CommandManager.cs
```diff
@@ -766,5 +779,25 @@ private void HandleNudge(ChatRoom room, ChatUser user, string[] parts)
...
+            _notificationService.ChangeNote(user);
+
+            _repository.CommitChanges();
+        }
+
+        private void HandleAfk(ChatUser user, string[] parts)
+        {
+            // First part is the command.
+            // Second is the AFK initial/prepending text.
+            var newParts = new List<string> {"afk", "Afk"};
```
davidfowl: Seems a little hacky

JabbR/Commands/CommandManager.cs
PureKrome: Easier?

``` javascript
 private void HandleNote(ChatUser user, string[] parts)
        {
            // We need to determine if we're either
            // 1. Setting a new Note.
            // 2. Clearing the existing Note.
            // If we have no optional text, then we need to clear it. Otherwise, we're storing it.
            bool isNoteBeingCleared = parts.Length == 1;
            user.Note = isNoteBeingCleared ? null : String.Join(" ", parts.Skip(1)).Trim();

            _notificationService.ChangeNote(user);

            _repository.CommitChanges();
        }
```

JabbR/ViewModels/UserViewModel.cs
PureKrome: I can ... but this is where it is mainly used:

Index.html / mobile.html

```
<span class="note{{if noteClass}} ${noteClass}{{/if}}" {{if note}}title="${note}"{{/if}}></span>
```

that line could/will get pretty messy....

JabbR/JabbR.csproj
PureKrome: Nope. manually fixed the csproj.

JabbR/Chat.ui.js
PureKrome: Fixed. I've never seen that operator before. Learnt something new this evening :)

JabbR/Hubs/Chat.cs
PureKrome: Good pickup. Fix0rd.

JabbR/Commands/CommandManager.cs
davidfowl: Might be simpler to not call HandleNode from HandleAfk.

JabbR/ViewModels/UserViewModel.cs
davidfowl: False, there's is js that is run before the template ever gets rendered. Just duplicate the logic in the client side js not in the template.

JabbR/Commands/CommandManager.cs
PureKrome: But what about DRY?

Afk is a note with some prepended text...

davidfowl: I'll sacrifice DRY any day for clarity. Plus it's only 3 lines.

PureKrome: here's the Afk code now. Yep, nicer.

``` javascript
private void HandleAfk(ChatUser user, string[] parts)
        {
            string message = String.Join(" ", parts.Skip(1)).Trim();
            user.Note = string.Format("{0} {1}", ChatUser.AfkPrependingText, message).Trim();

            _notificationService.ChangeNote(user);

            _repository.CommitChanges();
        }
```

davidfowl: Uppercase S for string.Format.

PR #264, opened 2011-12-20
- Missing Icon's.

PureKrome:
Damn ninja icons.
- Damn Clippy and StopWatch were being ninja's .. so I added them to the Project. _FacePlam_


davidfowl:
Already made the change :)

Issue #92, opened 2011-11-28
Add a /flag <country name> for your country :)


PureKrome:
Yo.

maybe have a tiny flag BEFORE the gravatar, on the left hand side.

also, autocomplete would be nice, for the country list ?

flags would be a single sprite .. and using css magical-unicorn-pewpew .. can display the correct one.

KThxBai !

/me hugs.

davidfowl:
You gonna contribute this?

PureKrome:
Hi David - sure :) Assign this bad boy to me.

What i was thinking was ..

/flag <[ISO 3166-1 alpha2](http://en.wikipedia.org/wiki/ISO_3166-1_alpha-2)>
/flag <no iso> removes any existing flag, if one was set.

ISO blah blah blah 2 == 2 char codes. US (for USA), AU (Australia), FR (France), etc...

For the ChatUser I was thinking of making this an ENUM (but that would a bit painful to make, the first time) or just a string. The DB will be an tinyint (i think they are under 255) or a smallint .. or char(2) if a string.

For the graphics, a [single image which will contain all the flags](http://flag-sprites.com/).  <-- ha! awesome .. already exists.

![flags](http://i.imgur.com/2T2VS.png)

Lastly, I suppose I was going to reference the gravatar / afk / note code to render the correct flag, if a flag has been set.

With Christmas here (in 1hr, 19 min...) i'll be afk for a few days as i'm wifed and kid'ed :~( but after that I can get straight onto it :)

davidfowl:
Is that image ok to use? (licensing etc)

Merry Christmas!

PureKrome:
I believe so. The license info is fond on the front page, right hand side .. towards the bottom..

> License
> FamFamFam flag icons are "available for free use for any purpose with no requirement for attribution"
> Blogpotato.de flag icons are licensed under Creative Commons
> Author and license terms for Maxmind icon set are unknown.

![](http://i.imgur.com/RjBFw.jpg)

PureKrome:
David, I'm doing a check on the user's input to make sure that the IsoCode they have provided is legit.

To do this, i'm basically grabbing the first 2 chars of their optional text. Then I was going to do an almighty switch statement .. like. ..

switch (isoCode)
{
case "au":
case "ab":
...
return;
default: throw new InvalidOperationException("blah blah iso code is invalid blah blah");
}

[...]

PR #280, opened 2011-12-30
Issue #92 - New /flag command.

PureKrome:
- Added new /flag command.
- Added unit tests to test the new command.
- Added DB migration for a new DB field.


JabbR/Services/ChatService.cs
```diff
@@ -490,5 +745,24 @@ internal static void ValidateNote(string note)
...
+            if (String.IsNullOrWhiteSpace(country))
+            {
+                throw new InvalidOperationException(
+                    "Sorry, but the country ISO code you requested doesn't exist. Please refer to http://en.wikipedia.org/wiki/ISO_3166-1_alpha-2 for a proper list of country ISO codes.");
+            }
+        }
+
+        internal static string GetCountry(string isoCode)
+        {
+            return String.IsNullOrWhiteSpace(isoCode) ||
```
davidfowl: Too much stuff in here, doesn't need to be this complicated. 

```
if(String.IsNullOrEmpty(isoCode)) {
    return null;
}

string country;
if(Countries.TryGetValue(isoCode, out country)) 
{
       return country;
}

return null;
```

JabbR/Commands/CommandManager.cs
```diff
@@ -809,5 +815,26 @@ private void HandleAfk(ChatUser user, string[] parts)
 
             _repository.CommitChanges();
         }
+
+        private void HandleFlag(ChatUser user, string[] parts)
+        {
+            string isoCode = String.Join(" ", parts.Skip(1)).Trim().ToLowerInvariant();
```
davidfowl: Don't do this. Just take parts[1] and do ToLowerInvariant

PureKrome: NP :)

JabbR/Services/ChatService.cs
PureKrome: Also NP :)

PR #282, opened 2012-01-02
Fixed issue 279 and also created cached jquery window and document objec...

redsquare:
...ts


JabbR/Chat.ui.js
```diff
@@ -653,24 +657,32 @@
...
                         ev.preventDefault();
                         return false;
+                    default:
+                        cyclingMessageHistory = false;
+
                 }
             });
 
+            function cycleMessage(messageHistoryDirection) {
+                var currentMessage = $newMessage[0].value;
```
davidfowl, line 675: Why not use `$newMessage.val()` ?

redsquare, line 675: Using jquery methods for everything is overhead especially when you are not worried about x-browser issues, in the same way I dont use .attr('id') or .attr('src') etc

davidfowl, line 675: Seems arbitrary. I don't think it's overhead for this case, what's the perf difference?

redsquare, line 675: Well its a good habit to get into I have found. See for yourself what the inbuilt val method actually does - http://james.padolsey.com/jquery/#v=git&fn=jQuery.fn.val. Why incur the extra cycles if you do not need to just so you use the library?

davidfowl, line 675: Because I really hate micro optimizations like this.

redsquare, line 675: Not sure I get how using a native js call is micro optimizing?

davidfowl, line 675: You're reason for using it is to avoid "jquery overhead", while everything else in the code we use jQuery to access things. This code path isn't even hot.

PR #293, opened 2012-01-12
Creation of UserVoice content provider

DavidChristiansen:
Hi guys,
Creation of a simple content provider for your consideration. This one is for uservoice articles.


DavidChristiansen:
Not to standard.

PR #300, opened 2012-01-13
Feature soundcloudcontentprovider

DavidChristiansen:
Hi guys,
Another content provider for your consideration. This time it's for soundcloud.com and the contentprovider spits out an HTML5 player.

As demonstrated here : http://jabbrnet.apphb.com/#/rooms/soundcloud

### DIFF
--- a/JabbR/ContentProviders/SoundCloudContentProvider.cs
+++ b/JabbR/ContentProviders/SoundCloudContentProvider.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using System.IO;
+using System.Net;
+using System.Text.RegularExpressions;
+using JabbR.ContentProviders.Core;
+using JabbR.Infrastructure;
+
+namespace JabbR.ContentProviders
+{
+    public class SoundCloudContentProvider : CollapsibleContentProvider
+    {
+
+        private static readonly Regex _titleRegex = new Regex("<meta.*content=\"(.*)\".*property=\"og:title\".*/>");
+        private static readonly Regex _trackIDExtractRegex = new Regex("<meta.*content=\"http://player\\.soundcloud\\.com/player\\.swf.*tracks%2F(.*?)&amp;.*property=\"og:video\"\\s*/>");
+
+        protected override ContentProviderResultModel GetCollapsibleContent(HttpWebResponse response)
+        {
+            using (var sr = new StreamReader(response.GetResponseStream()))
+            {
+                var pageContent = sr.ReadToEnd();
+                if (String.IsNullOrEmpty(pageContent))
+                {
+                    return null;
+                }
+
+                var trackID = _trackIDExtractRegex.SingleOrDefault(pageContent);
+                var titleContent = _titleRegex.SingleOrDefault(pageContent);
+
+                if (trackID == null || titleContent == null)
+                {
+                    return null;
+                }
+
+                return new ContentProviderResultModel() {
+                    Title = titleContent,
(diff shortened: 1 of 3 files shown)


JabbR/ContentProviders/SoundCloudContentProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+			return new ContentProviderResultModel() 
+			{
+				Title = article.title,
+				Content = article.html
+			};
+		}
+
+		private static dynamic FetchArticle(Uri url)
+		{
+			var webRequest = (HttpWebRequest)WebRequest.Create(String.Format(_soundcloudAPIURL, url.AbsoluteUri));
```
sethwebster: Is it possible to do this without the second request?

DavidChristiansen: Let me look into this further.

DavidChristiansen: Yes, it is

davidfowl:
Looks like you're using tabs.

JabbR/ContentProviders/SoundCloudContentProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+
+		private static readonly Regex _titleRegex = new Regex("<meta.*content=\"(.*)\".*property=\"og:title\".*/>");
+		private static readonly Regex _trackIDExtractRegex = new Regex("<meta.*content=\"http://player\\.soundcloud\\.com/player\\.swf.*tracks%2F(.*?)&amp;.*property=\"og:video\"\\s*/>");
+
+		protected override ContentProviderResultModel GetCollapsibleContent(HttpWebResponse response)
+		{
+			using (var sr = new StreamReader(response.GetResponseStream())) 
+			{
+				var pageContent = sr.ReadToEnd();
+				if (string.IsNullOrEmpty(pageContent))
```
davidfowl: No self closing ifs add braces.
Using String. not string.

JabbR/ContentProviders/SoundCloudContentProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+					.Groups
+					.Cast<Group>()
+					.Skip(1)
+					.Select(g => g.Value).SingleOrDefault(v => !String.IsNullOrEmpty(v));
+				
+				var titleContent = _titleRegex.Match(pageContent)
+					.Groups
+					.Cast<Group>()
+					.Skip(1)
+					.Select(g => g.Value).SingleOrDefault(v => !String.IsNullOrEmpty(v));
```
davidfowl: Why do you have the same code 2 times in a row. Actually I find it strange that nobody has created a helper for this yet. How many times is this code repeated in content providers.

JabbR/ContentProviders/SoundCloudContentProvider.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Text.RegularExpressions;
+using JabbR.ContentProviders.Core;
+using Newtonsoft.Json;
+
+namespace JabbR.ContentProviders
+{
+	public class SoundCloudContentProvider : CollapsibleContentProvider
+	{
+
+		private static readonly Regex _titleRegex = new Regex("<meta.*content=\"(.*)\".*property=\"og:title\".*/>");
```
davidfowl: Don't like verbatim strings?

DavidChristiansen:
Was written on my work PC so maybe the setting were asque

JabbR/ContentProviders/SoundCloudContentProvider.cs
DavidChristiansen: From the point of view of futureproofing, I don't want the regex to break should the spacing within this meta tag change. The regex is told to care about the three keypoints that are important - the tag meta, the attribute content and the property of og:title, everything else can change and the regex wont care.

DavidChristiansen:
Please ignore last checkin - Something going wrong on the PC i was using and it is messing up either line ending or spacing.

davidfowl:
Looks good in general but the braces are messed up.

PR #301, opened 2012-01-13
Bash.org content provider

phillip-haydon:
Added a new content provider to grab quotes from bash.org
Added the HTMLAgilityPack to scrape HTML

### DIFF
--- a/JabbR/ContentProviders/BashQDBContentProvider.cs
+++ b/JabbR/ContentProviders/BashQDBContentProvider.cs
@@ -1,0 +1,75 @@
+﻿using System;
+using System.Linq;
+using System.Net;
+using System.Text.RegularExpressions;
+using HtmlAgilityPack;
+using JabbR.ContentProviders.Core;
+
+namespace JabbR.ContentProviders
+{
+    public class BashQDBContentProvider : CollapsibleContentProvider
+    {
+        private static readonly Regex QueryRegex = new Regex(@"\d+");
+        private static readonly string ContentFormat = @"
+<div class=""bashqdb_wrapper"">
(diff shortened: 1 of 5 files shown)


JabbR/ContentProviders/BashQDBContentProvider.cs
```diff
@@ -0,0 +1,75 @@
...
+            using (var responseStream = response.GetResponseStream())
+            {
+                var htmlDocument = new HtmlDocument();
+                htmlDocument.Load(responseStream);
+                htmlDocument.OptionFixNestedTags = true;
+
+                var quote = htmlDocument.DocumentNode
+                                        .SelectSingleNode("//body")
+                                        .SelectNodes("//p").Where(a => a.Attributes.Any(x => x.Name == "class" && x.Value == "qt"))
+                                        .Single().InnerHtml;
```
davidfowl: You might want to code more defensively here.

JabbR/ContentProviders/BashQDBContentProvider.cs
```diff
@@ -0,0 +1,95 @@
...
+        <a href=""{0}"" target=""_blank"">{1}</a>
+    </div>
+    <div class=""bashqdb_content"">{2}</div>
+</div>";
+
+        private static readonly string[] WhiteListHtml = new[] {"br", "#text"};
+
+        protected override ContentProviderResultModel GetCollapsibleContent(HttpWebResponse response)
+        {
+            var pageInfo = ExtractFromResponse(response);
```
davidfowl, line 23: What happens if you return null here? I don't see a check.

Issue #190, opened 2011-12-07
Need a way to close a room


haacked:
If you create a locked room, you can never unlock it (privacy concerns). So that room name is effectively locked forever. Would be nice to be able to close a room so the name frees up.

PureKrome:
Out of interest .. should this then delete the entire history of the room and the room itself from the DB?

If not .. (say .. for archiving purposes) ... 
- How would anyone get access to it? Only via the DB?
- Do we care about archiving?

PR #303, opened 2012-01-14
Issue #190 - Close a room

PureKrome:
So I've given this a crack.

new command:

/close <room name>

A room now has a new property => `bool IsOpen`. My migration (`201201130555046_RoomIsOpen.cs`) hopefully sets it by default (mainly for existing rooms) to TRUE .. meaning all rooms are open, initially. By default new rooms created are also open.

``` c#
public ChatRoom()
{
    ....
    IsOpen = true;
}
```

Closing a room just sets the property `IsOpen = false` and then i persist it to the store.

Two of the important structural changes have been 

``` c#
IJabbrRepo

ChatRoom GetRoomByNameAndIsOpen(string roomName, bool isOpen);
```

So it will retrieve a room by name AND the IsOpen status.

``` c#
ChatRoom GetRoomByName(string roomName);
```

that remains untouched, still.

Next, the `VerifyRoom` has been changed to this...

``` c#
public static ChatRoom VerifyRoom(this IJabbrRepository repository, string roomName)
{
    return VerifyRoom(repository, roomName, true);
}

public static ChatRoom VerifyRoom(this IJabbrRepository repository, string roomName, bool? isOpen)
{ ... }
```

So this means the code is backwards compatible AND all existing code therefore requires a room to exist, etc.. BUT also be Open (ie. the IsOpen == true, in the third paramater, above).

That's really about it :)

Looking forward to seeing what you gents think.

OH! I'll need to create a new issue, after this. a /reopen command .. to open an existing closed room. (unless u want ADMIN to do that, for the once in a million times that's required.

A few side notes:-
1. You cannot close a room if there's people in it.
2. A room can only be closed by a person who is an owner. Remember, rooms can have multiple owners, which fine :)
3. The room is not shown in the Lobby if it's closed. It still exists in the DB.
4. Trying to join a closed room == 'Unable to find room 'qwerty''


davidfowl:
Not a fan of IsOpen, I'd prefer a Closed bit.

JabbR/Models/InMemoryRepository.cs
```diff
@@ -82,9 +82,21 @@ public ChatRoom GetRoomByName(string roomName)
             return _rooms.FirstOrDefault(r => r.Name != null && r.Name.Equals(roomName, StringComparison.OrdinalIgnoreCase));
         }
 
+        public ChatRoom GetRoomByNameAndIsOpen(string roomName, bool isOpen)
```
davidfowl: Why do you need this on the repository?

JabbR/Services/ChatService.cs
```diff
@@ -735,6 +735,38 @@ public void LockRoom(ChatUser user, ChatRoom targetRoom)
...
+            }
+
+            // Make sure there's no one in the room.
+            if (targetRoom.Users != null && targetRoom.Users.Count > 0)
+            {
+                throw new InvalidOperationException(
+                    String.Format("Room '{0}' has {1} user{2} still in it. Unable to close a room while users are still in it",
+                        targetRoom.Name, 
+                        targetRoom.Users.Count, 
+                        targetRoom.Users.Count == 1 ? string.Empty : "s"));
```
davidfowl: String.Empty not string.Empty.

JabbR/Services/ChatService.cs
```diff
@@ -735,6 +735,38 @@ public void LockRoom(ChatUser user, ChatRoom targetRoom)
...
+        {
+            EnsureOwner(user, targetRoom);
+
+            if (!targetRoom.IsOpen)
+            {
+                throw new InvalidOperationException(String.Format("{0} is already closed.", targetRoom.Name));
+            }
+
+            // Make sure there's no one in the room.
+            if (targetRoom.Users != null && targetRoom.Users.Count > 0)
```
davidfowl: You want to check only online users not users in general. People are always in room unless they explicitly leave. If you close a room that someone is in you should remove everyone from that room.

JabbR/Services/ChatService.cs
```diff
@@ -632,7 +632,7 @@ private static void EnsureOwner(ChatUser user, ChatRoom room)
         {
             if (!room.Owners.Contains(user))
             {
-                throw new InvalidOperationException("You are not an owner of " + room.Name);
+                throw new InvalidOperationException("You are not an owner of room '" + room.Name + "'");
```
davidfowl, line 635: Why?

JabbR/Models/RepositoryExtensions.cs
```diff
@@ -63,6 +63,11 @@ public static ChatUser VerifyUserId(this IJabbrRepository repository, string use
 
         public static ChatRoom VerifyRoom(this IJabbrRepository repository, string roomName)
         {
+            return VerifyRoom(repository, roomName, true);
+        }
+
+        public static ChatRoom VerifyRoom(this IJabbrRepository repository, string roomName, bool? isOpen)
```
davidfowl: Why is this nullable?

JabbR.Test/CommandManagerFacts.cs
```diff
@@ -3200,7 +3406,7 @@ public void NoIsoCodeClearsFlag()
             public void IncorrectIsoCodeThrows()
             {
                 // Arrange.
-                const string isoCode = "xx";
+                const String isoCode = "xx";
```
davidfowl: Why is this uppercase? The only thing that is uppercase is method calls not declarations. Look at the rest of the code.

JabbR/Models/InMemoryRepository.cs
PureKrome: Because later on, when we `VerifyRoom` .. I wish to make sure a verified room includes rooms that are only Open. But it's possible that we will want (later on) the details of a room, regardless of it's open/closed state .. say for admin stuff.

JabbR/Services/ChatService.cs
PureKrome: Doh. yep, np.

JabbR/Models/InMemoryRepository.cs
davidfowl: Forget about later on.

JabbR/Services/ChatService.cs
PureKrome: Ah, ok. That I didn't know. Ok, so if i'm Idle, am I also online? Does `LastActivity` determine if a person is idle or not, while `ConnectedClients` lists all the user clients (eg. ipad, browsers, etc) which a person is `currently` connected to JabbR?

JabbR/Services/ChatService.cs
PureKrome, line 635: Do you , why did i elaborate on the exception message? Because i wanted to standardize the messages sent to the user about rooms. When i was testing, I found the messages were 'poor'.

Eg. You are not the owner of cricket

Just didn't sound as nice as 'You are not the owner of the _room_ 'cricket'' (emphasis mine)

JabbR/Services/ChatService.cs
davidfowl: Just use the Online extension method. Idle is still online.

JabbR/Models/RepositoryExtensions.cs
PureKrome: Because it's possible that you might not need to verify a room with the open/closed state. I'm thinking this might be the case for admin stuff. While we don't have it yet, I might have been getting ahead of myself. Simple to force it to NOT be nullable and if (when?) admin stuff gets added, then deal with it then.

davidfowl: Do it when we get to it, YAGNI.

JabbR.Test/CommandManagerFacts.cs
PureKrome: I thought -all- strings should be Strings. eg. var foo = new String.Split(..);

PureKrome:
Doh. i originally started out (and even commited) the property to be `IsClosed`. I found that i started having a lot more `!IsClosed` in my logic .. instead of just `if (IsOpen)` .. so i thought it read nicer. If you want, i'll change over ... so you prefer me to change it over?

JabbR/Models/InMemoryRepository.cs
PureKrome: Rgr Rgr.

JabbR/Models/RepositoryExtensions.cs
PureKrome: Awesome :) Can do, sir!

JabbR.Test/CommandManagerFacts.cs
davidfowl: method calls.

davidfowl:
Yes and call it Closed not IsClosed.

JabbR.Test/CommandManagerFacts.cs
PureKrome: Cheers - gotcha (and sorry about that) _blush_.

PureKrome:
Sure thing! on it now :)

JabbR/Models/RepositoryExtensions.cs
PureKrome: Also, are you happy with me leaving the original method there, which will call my new one (but `bool? isOpen` will be refatored to `bool closed`) ?

Why - there's a number of places that call `VerifyRoom`, such as `HandleAllow`, `HandleAddOwner`, etc. If i leave the original method there, then it's assumed that all those existing methods require the room to be open.

Otherwise, I'll merge both of those methods and then update -all- references to the now defunct method, adding in the new `false` argument.

eg. 

``` c#
private void HandleAddOwner(ChatUser user, string[] parts)
{
    ....
    ChatRoom targetRoom = _repository.VerifyRoom(roomName, false);
}
```

Just not sure if having two methods is not what you want and you don't mind me updating all those other methods.

PureKrome: Bleh - Please forget that last post. Totally not required. _blush_ To `VerifyRoom` we always require the room to be Open. Simple. extra params or methods or crap. 

Sorry about that.

JabbR/Services/ChatService.cs
PureKrome: Ah! found it :) awesome. did not know that existed _blush_

much better :)

``` C#
if (targetRoom.Users != null && targetRoom.Users.Online().Any())
```

JabbR/Models/ChatRoom.cs
```diff
@@ -32,6 +33,7 @@ public ChatRoom()
             Messages = new HashSet<ChatMessage>();
             Users = new HashSet<ChatUser>();
             AllowedUsers = new HashSet<ChatUser>();
+            Closed = false;
```
davidfowl: There's no need to set it to false. The CLR will do this for you :)

JabbR/Models/PersistedRepository.cs
```diff
@@ -80,8 +80,11 @@ public ChatRoom GetRoomByName(string roomName)
 
         public IQueryable<ChatRoom> GetAllowedRooms(ChatUser user)
         {
-            // All public rooms and private rooms the user can see
-            return _db.Rooms.Where(r => !r.Private || r.Private && r.AllowedUsers.Any(u => u.Key == user.Key));
+            // All *open* public and private rooms the user can see.
+            return _db.Rooms
```
davidfowl, line 84: Should the owner be able to see the closed room?

JabbR/Services/ChatService.cs
```diff
@@ -735,6 +735,39 @@ public void LockRoom(ChatUser user, ChatRoom targetRoom)
...
+                        targetRoom.Name,
+                        onlineUsers,
+                        onlineUsers == 1 ? String.Empty : "s"));
+            }
+
+            // Make the room closed.
+            targetRoom.Closed = true;
+
+            // Make all users in the current room allowed
+            foreach (var u in targetRoom.Users.Online())
```
davidfowl: Why do you add allowed users on close?

davidfowl:
Try running this code locally on rooms with alot of people. Generate some fake data and test it out.

JabbR/Models/PersistedRepository.cs
PureKrome, line 84: I thought 'no'. If there's a problem, let them hassle you. Basically, YAGANI.

JabbR/Services/ChatService.cs
PureKrome: <del>I don't see where I'm doing that, in the code snippet, above.</del>


Ahh ... very bad copy/paste from the lock code, above. Will remove.

PureKrome:
Rgr. will do tomorrow at work :) thanks for being patient, David :)

JabbR/Models/ChatRoom.cs
PureKrome: Fixed. 

(that was a hangup from having IsOpen = true, originally).

JabbR/Services/ChatService.cs
PureKrome: Fixed.

JabbR/Services/ChatService.cs
```diff
@@ -735,6 +735,33 @@ public void LockRoom(ChatUser user, ChatRoom targetRoom)
...
+        {
+            EnsureOwner(user, targetRoom);
+
+            if (targetRoom.Closed)
+            {
+                throw new InvalidOperationException(String.Format("{0} is already closed.", targetRoom.Name));
+            }
+
+            // Make sure the (owner) user is not in the room.
+            if (targetRoom.Users.Any(x => x.Id == user.Id))
```
davidfowl: Contains?

JabbR/Hubs/Chat.cs
```diff
@@ -643,7 +664,10 @@ void INotificationService.ShowUserInfo(ChatUser user)
             Caller.showUserInfo(new
             {
                 Name = user.Name,
-                OwnedRooms = user.OwnedRooms.Allowed(userId).Select(r => r.Name),
+                OwnedRooms = user.OwnedRooms
+                    .Allowed(userId)
+                    .Where(x => !x.Closed)
```
davidfowl: `r => r` not x

PR #328, opened 2012-01-27
In order to fix my error I implemented extensible client side scripting

ToJans:
This pull request adds extensible client scripting to Jabbr; it intercepts the message before it is sent.

This adds the following possibilities:

Custom client actions:

 Try entering "/alert blah"; it will intercept the message and show an alert box with "blah"

Load external plugins from a gist, f.e. https://gist.github.com/1688574 :

 Enter "/gistplugin 1688574"

After loading this gistplugin, every single "zomg" in a message will be replaced by the
  full length string "Zombies Oh My Gawd !!!"

I hope this improves my reputation a bit ;)

@ToJans

### DIFF
--- a/JabbR/Chat.ClientsideHooks.js
+++ b/JabbR/Chat.ClientsideHooks.js
@@ -90,8 +90,6 @@
         });
     }
 
-
-
     return {
         plugins: plugins,
         interceptMessage: interceptMessage,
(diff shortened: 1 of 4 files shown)


ToJans:
For the record: this time I tested it quite thoroughly ;)

davidfowl:
What's this even for? Testing?

ToJans:
I think the best description might be "client side Hubot" ;)

ToJans:
FYI, I would not feel offended if this pull request would be refused.

Issue #287, opened 2012-01-09
Keep focus when changing tabs


dannylloyd:
When switching tabs, the focus is lost from the textbox. I don't know if it's just me, but I like to switch tabs and begin typing, I have switch tab then click on the textbox to type a message.

DamianEdwards:
Agreed, particularly annoying when using keyboard shortcuts to move to
new tabs.

Sent from my Windows Phone
From: dannylloyd
Sent: 1/9/2012 10:46 AM
To: DamianEdwards
Subject: [JabbR] Keep focus when changing tabs (#287)
When switching tabs, the focus is lost from the textbox. I don't know
if it's just me, but I like to switch tabs and begin typing, I have
switch tab then click on the textbox to type a message.

---

Reply to this email directly or view it on GitHub:
https://github.com/davidfowl/JabbR/issues/287

dannylloyd:
Off-topic, but where can I find the keyboard shortcuts?

DamianEdwards:
I just meant the accesskey attribute which we populate for a bunch of
links, like the tab headers.

Sent from my Windows Phone

dannylloyd:
Cool, those are nice. It makes this focus feature all the more important.

dannylloyd:
There is an accesskey for the message textbox, "m".

davidfowl:
This is still broken for ipads. They aren't detected properly and it makes the keyboard show up whenever rooms are switched.

Issue #288, opened 2012-01-11
/help doesn't mention #roomname


dannylloyd:
/help lists a lot of features, but it doesn't mention this one.

PR #329, opened 2012-01-28
Added focus for issue #287

JesseBuesking:
Nothing crazy, but this is my first pull request for an open source project, ever, so if I'm doing something crazy just tell me :D

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -181,6 +181,9 @@
             this.users.addClass('current')
                       .show();
 
+            // setting focus to the message box #287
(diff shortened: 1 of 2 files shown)


JesseBuesking:
Just updated Chat.cs for issue #288, not realizing it'd be added to this pull request.

davidfowl:
This change is great except for when you use it on an iPad, iPhone, any mobile device?. You should check for that since auto-focus on the textbox causes the keyboard to popup.

davidfowl:
I'd start over this pull request. Always do separate features in a branch not on top of each other.

Issue #281, opened 2012-01-02
New tabs created automatically


dskc:
I've created a room with the title as ".NETv4.5". This appears in the lobby list. When I click on that, I see a tab on the top of the body chat window and I'm still at the lobby.

When i click on the ".NETv4.5" tab, it automatically created one more tab with the same title. It continues as long as you click and you endup with as many tabs as you have pressed. But the "x" close button doesn't work for these newly created tabs.

If anyone is not able to reproduce this issue, please revert me at my personal mail id, VARTHY [at] GMAIL [at] COM, so that I can demostrate them the problem.

Thanks for reading.

PR #330, opened 2012-01-28
Fix for issue #281

dpen2000:
The original issue reproduces with any room name with two "." in the name. JQuery seems to have trouble selecting id's with these in using the normal "#the.i.d" selector syntax: switching to the "[id="the.i.d"]" syntax works fine and fixes issue.

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -212,9 +212,9 @@
 
     function getRoomElements(roomName) {
         var roomId = getRoomId(roomName);
-        return new Room($('#tabs-' + roomId),
-                        $('#users-' + roomId),
-                        $('#messages-' + roomId));
+        return new Room($('[id="tabs-' + roomId + '"]'),
+                        $('[id="users-' + roomId + '"]'),
+                        $('[id="messages-' + roomId + '"]'));
     }
 
     function getCurrentRoomElements() {


JabbR/Chat.ui.js
```diff
@@ -212,9 +212,9 @@
 
     function getRoomElements(roomName) {
         var roomId = getRoomId(roomName);
-        return new Room($('#tabs-' + roomId),
-                        $('#users-' + roomId),
-                        $('#messages-' + roomId));
+        return new Room($('[id="tabs-' + roomId + '"]'),
```
davidfowl, line 215: Does this slow anything down? Does jQuery optimize #blah vs [id=blah]?

davidfowl, line 215: This we should only go the slow path if the id contains funny chars.

phillip-haydon:
IMO we shouldn't allow room names to contain non-alpha/numeric characters. 
a-z, 0-9, no spaces/special characters.

davidfowl:
+1, we can just update the logic in [IsValidRoomName](https://github.com/davidfowl/JabbR/blob/429c554e46bd971b6c31a51cdc6c75e1a86151ee/JabbR/Services/ChatService.cs#L674). We just won't allow . anymore maybe - is ok.

phillip-haydon:
And -, since the general chat room wouldn't work anymore without it lol.

Issue #314, opened 2012-01-25
Support GitHub / Campfire Emojis in Jabbr


anaisbetts:
All of the Emojis from http://www.emoji-cheat-sheet.com/ should work in Jabbr. Because I need my :tada:s and my :sparkles:
## Wtf how am I supposed to do this? I'm extremely lazy

https://api.github.com/emojis  - just download this list from time to time and sub in the Emoji tag for the image. Bam!  :muscle: :+1: :kissing_face:

PR #333, opened 2012-01-28
Added support for Emojis - Issue #314

TrentScholl:
Add support for Emojis using a new static method in the TextTransform class.

Works in public and private messages.


davidfowl:
Wish we had spoken about it before you did the work. We decided yesterday in chat that this should be done client side.

TrentScholl:
All good.

I'll update it

davidfowl:
Also Philliphaydon was gonna work on this so you guys should sync up.

Sent from my Windows Phone

PR #335, opened 2012-01-29
added focus event on tab switch for issue #287

JesseBuesking:
Created a branch and fix for the focus event.  Added a utility variable for checking if a browser is mobile-based, which is used to determine whether to set focus.


JabbR/Chat.ui.js
```diff
@@ -165,6 +165,8 @@
             var currUnread = this.getUnread(),
                 lastUnread = this.messages.find('.message-separator').data('unread') || 0;
 
+            !utility.isMobile && $newMessage.focus();
```
davidfowl: Whats this? Where's the if statement

JesseBuesking: That is the equivalent of

``` js
if (!utility.isMobile) {
    $newMessage.focus();
}
```

Maybe it's a bit difficult to read?  If you want I can just do the simple `if` block.  Your call :D

davidfowl: Yes please, and merge with upstream.

Issue #286, opened 2012-01-08
Add search for rooms in "Lobby"


jasonsirota:

jasonsirota:
Browser search works in the meantime so can be closed if not a priority.

dpen2000:
I'll do this.

PR #338, opened 2012-01-29
#286 - add search for rooms in "lobby"

dpen2000:
Added search for rooms based on John Resig's LiveSearch with a few modifications.


JabbR/Chat.css
```diff
@@ -33,6 +33,14 @@
     height: 18px;
 }
 
+#users-filter {
```
davidfowl, line 36: Fix formatting.

JabbR/Chat.ui.js
```diff
@@ -24,7 +24,9 @@
         name,
         lastCycledMessage = null,
         $window = $(window),
-        $document = $(document);
+        $document = $(document),
+        roomFilter = null,
```
davidfowl: All the other jQuery variables start with $

JabbR/Chat.ui.js
```diff
@@ -159,6 +161,10 @@
 
             this.users.removeClass('current')
                       .hide();
+
+            if (this.isLobby()) {
```
davidfowl, line 164: Is this supposed to be `if(!this.isLobby())`

JabbR/index.htm
```diff
@@ -198,6 +200,9 @@
         <div id="chat-area">
             <ul id="messages-lobby" class="messages current">
             </ul>
+            <form id="users-filter-form" action="#">
+			<input id="users-filter" class="filter" type="text" />
```
davidfowl: Formatting here as well.

JabbR/Chat.ui.js
dpen2000, line 164: When we're moving away from lobby, in this function, isLobby returns true and so we hide the textbox for filtering rooms.

JabbR/Chat.ui.js
```diff
@@ -846,6 +856,8 @@
                     $li.addClass('locked');
                 }
             });
+            ui.$roomFilter.update();
+            $roomFilterInput.val('');
```
davidfowl, line 863: Why is this in run? What does it do

JabbR/Chat.ui.js
```diff
@@ -741,6 +750,7 @@
                 }
             },
             { unescape: ',/' });
+            ui.$roomFilter = $roomFilterInput.liveUpdate('#users-lobby', true);
```
dpen2000: This is the line in Run. This is initializing the LiveUpdate plugin which is powering the room search. I think this should probably be moved to the initialize function.

JabbR/Chat.ui.js
dpen2000, line 863: These two lines are in populateLobbyRooms: I'm updating the LiveUpdate plugin's cache of items (based on the items that have just been populated) and setting the filter back to empty.

davidfowl, line 863: Maybe add some comments.

JabbR/Chat.ui.js
davidfowl: me too :). And add some comments.

Scutua (approved):

PR #342, opened 2012-01-30
Creation of nuget package content provider

DavidChristiansen:
Hi guys,
Here's a NuGet package content provider for your consideration. 
DC

### DIFF
--- a/JabbR/ContentProviders/NugetNuggetContentProvider.cs
+++ b/JabbR/ContentProviders/NugetNuggetContentProvider.cs
@@ -13,12 +13,8 @@ namespace JabbR.ContentProviders
     public class NugetNuggetContentProvider : CollapsibleContentProvider
     {
         private static readonly Regex _nugetURLRegex = new Regex(@"nuget\.org/packages/(.*)");
-
-        private static readonly string _nugetFeedURL =
-            "http://packages.nuget.org/v1/FeedService.svc/Packages()?$filter=Title eq '{0}'&$orderby=Created desc";
-
-        private static readonly string _nugetBadgeFormat =
-            "<div class=\"nuget-badge\"><code>PM> Install-Package {0}</code><div class=\"nuget-description\">{1}{2}</div><div class=\"nuget-projectinfo\">{3}</div><div style=\"clear:both\"></div></div>";
+        private static readonly string _nugetFeedURL = "http://packages.nuget.org/v1/FeedService.svc/Packages()?$filter=Title eq '{0}'&$orderby=Created desc";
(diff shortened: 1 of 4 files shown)


JabbR/ContentProviders/NugetNuggetContentProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+                                   Title = packageInfo.Title + " NuGet package"
+                               };
+                }
+            }
+            return null;
+        }
+
+        private dynamic FetchPackage(string packageName)
+        {
+            var webRequest = (HttpWebRequest) WebRequest.Create(String.Format(_nugetFeedURL, packageName));
```
sethwebster, line 65: Can this be done without a second request?

JabbR/ContentProviders/NugetNuggetContentProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        private static readonly string _nugetBadgeFormat =
+            "<div class=\"nuget-badge\"><code>PM> Install-Package {0}</code><div class=\"nuget-description\">{1}{2}</div><div class=\"nuget-projectinfo\">{3}</div><div style=\"clear:both\"></div></div>";
+
+        protected override ContentProviderResultModel GetCollapsibleContent(HttpWebResponse response)
+        {
+            string packageName = ExtractPackageName(response.ResponseUri);
+            if (!String.IsNullOrEmpty(packageName))
+            {
+                var package = FetchPackage(packageName);
```
sethwebster: Maybe use dynamic here to avoid needing the ugly cast?

JabbR/ContentProviders/NugetNuggetContentProvider.cs
DavidChristiansen, line 65: I can't obtain the information I want from the page itself without some awful html parsing. I considered this the best solution and no different to what's done in existing content providers.

JabbR/ContentProviders/NugetNuggetContentProvider.cs
DavidChristiansen: Open to suggestions on how to otherwise obtain the length of an array from a dynamic typed response from JSONConvert.

JabbR/ContentProviders/NugetNuggetContentProvider.cs
sethwebster, line 65: Some of the other providers have been refactored, changed, to remove 2nd request when it wasn't really needed.  It's a default question to ensure folks are thinking through the fact that you already have made and have an open request with which to work.

JabbR/ContentProviders/NugetNuggetContentProvider.cs
sethwebster: I may be not thinking this thru, and it's really not that big of a deal but, you can use dynamic as seen in another provider using the same approach.

if you simply use the dynamic keyword instead of var, you can reference properties/methods simply by knowing they are there.

DavidChristiansen: done

PR #350, opened 2012-02-01
Move scrollbar position to top when user types in room filter

dpen2000:
This is nicer behaviour.


davidfowl:
Works in all browsers?

dpen2000:
Tested and works on Windows in IE 9, Firefox 9+10, Chrome 16 and Safari 5.

PR #361, opened 2012-02-03
Change in css to make large images look properly in the chatbox.

ignaciofuentes:
Basically all I did was add a max-width to the images in the css.


davidfowl:
Max width based on what?

ignaciofuentes:
based on the h3 on top of the image.... I have posted an image on meta room that shows the error... how else would you fix it?

davidfowl:
How does it work on high resolutions, is it always a fixed size?

PR #365, opened 2012-02-04
Issue 306

dpen2000:
This is the /open command.

### DIFF
--- a/JabbR/Services/ChatService.cs
+++ b/JabbR/Services/ChatService.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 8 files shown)


JabbR/Services/ChatService.cs
```diff
@@ -810,6 +810,20 @@ public void CloseRoom(ChatUser user, ChatRoom targetRoom)
             _repository.CommitChanges();
         }
 
+        public void OpenRoom(ChatUser user, ChatRoom targetRoom)
+        {
+            EnsureOwner(user, targetRoom);
+
+            if (targetRoom.Closed == false)
```
davidfowl: `if(!targetRoom.Closed)`

PR #371, opened 2012-02-12
Implementation of /clearmessages for room owners to reset history

paultyng:
Added a way for owners to clear the history of the room.


davidfowl:
2 things:
1. I don't even like this feature.
2. If I did like it, it'd be an admin feature.

Issue #372, opened 2012-02-12
Ability to download / export room messages


paultyng:
Main requirements captured from chat:
## Server Implementation
- `IHttpHandler` allowing users to download chat history for a room
- User must be logged in and in room if its a locked room
- GET request with query string params for format and date range in server timezone
- For public requests data should be available for possible mashup on other sites
## User Interface
- [Icon](http://thenounproject.com/noun/download/#icon-No435) next to the popup notifications icon
- Clicking brings up twitter bootstrap dialog with options:
  - Date range preset (last hour, last day, last week, last month, all) (no custom option yet due to timezone issues)
  - Possibly format options: json, csv, xml (first implementation in JSON?)

For the private rooms, it may be possible to grab history from a site with XSRF, so not sure if we want to limit private room downloads to POST or something?

PR #374, opened 2012-02-12
Room message download feature

paultyng:
#372


paultyng:
Submitted an additional commit for the CSS file, that was set to tabs not spaces.

JabbR/Handlers/DownloadHandler.cs
```diff
@@ -0,0 +1,117 @@
...
+using Newtonsoft.Json;
+using System.Globalization;
+
+namespace JabbR.Handlers
+{
+    public class DownloadHandler : IHttpHandler
+    {
+        const string FilenameDateFormat = "yyyy-MM-dd.HHmmsszz";
+
+        IJabbrRepository _repository;
```
davidfowl, line 15: add private modifier.

davidfowl:
Also, merge with the latest.

PR #378, opened 2012-02-13
Reducing the craziness in SoundCloud provider (don't Regex All the Things)

alexkopy:
Old Way:
1) Get permalink URL from user
2) Download the target HTML page
3) Grep for ID
4) Manually build iframe html

New Way:
1) Get permalink URL from user
2) Query SoundCloud's OEmbed Endpoint with that URL
3) Return Title and iframe code, nicely provided in JSON response from SoundCloud
4) Profit!

### DIFF
--- a/JabbR/ContentProviders/SoundCloudContentProvider.cs
+++ b/JabbR/ContentProviders/SoundCloudContentProvider.cs
@@ -1,42 +1,59 @@
 ﻿using System;
 using System.IO;
-using System.Linq;
 using System.Net;
-using System.Text.RegularExpressions;
+using System.Threading.Tasks;
 using JabbR.ContentProviders.Core;
-using JabbR.Infrastructure;
+using Newtonsoft.Json;
 
 namespace JabbR.ContentProviders
 {
     public class SoundCloudContentProvider : CollapsibleContentProvider
     {
-        private static readonly Regex _titleRegex = new Regex("<meta.*content=\"(.*)\".*property=\"og:title\".*/>");
-        private static readonly Regex _trackIDExtractRegex = new Regex("<meta.*content=\"http://player\\.soundcloud\\.com/player\\.swf.*tracks%2F(.*?)&amp;.*property=\"og:video\"\\s*/>");
+        private sealed class SoundCloudResponse
+        {
+            [JsonProperty(PropertyName = "title")]
+            public string Title { get; set; }
+            [JsonProperty(PropertyName = "html")]
+            public string FrameMarkup { get; set; }
+        }
 
         protected override ContentProviderResultModel GetCollapsibleContent(HttpWebResponse response)
         {
-            using (var sr = new StreamReader(response.GetResponseStream()))
-            {
-                var pageContent = sr.ReadToEnd();
-                if (String.IsNullOrEmpty(pageContent))
-                {
-                    return null;
-                }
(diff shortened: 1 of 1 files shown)


davidfowl:
2 questions:
1. What does this do that's better?
2. Why are you using making the request using a task?

alexkopy:
@davidfowl 
1) Are you saying you don't see anything wrong with downloading an entire webpage (intended to be presented to the user, thus including everything), and greping over it (twice) to get what you want, when there is api available for the exact purpose? Memory and speed savings are the wins here.
2) Just following standard practice in JabbR, that is how this is handled upstream (see ResourceProcessor.ExtractResource). The entire Provider architecture is based on a flawed assumption that the URL that the user supplies is the one we will always fetch. In case of SoundCloud, user's URL is a parameter, so I have to redo the work that was done in ExtractResource (you'll see I do the same thing).

davidfowl:
Memory and speed savings seriously? Did you measure anything? 

Anyways, I'm not saying that you're change ins't good I just wanted know if you were just re-factoring and having it work the same way.

One thing we could do is make the external web request Lazy so content providers that don't need it don't end up making the initial request.

JabbR/ContentProviders/SoundCloudContentProvider.cs
```diff
@@ -1,47 +1,64 @@
...
-
-                var trackID = _trackIDExtractRegex.FindMatches(pageContent).SingleOrDefault();
-                var titleContent = _titleRegex.FindMatches(pageContent).SingleOrDefault();
-
-                if (trackID == null || titleContent == null)
-                {
-                    return null;
-                }
+    {
+        private sealed class SoundCloudResponse
```
davidfowl: Move this to the bottom of the file.

JabbR/ContentProviders/SoundCloudContentProvider.cs
```diff
@@ -1,47 +1,64 @@
...
+                                            var json = reader.ReadToEnd();
+                                            widgetInfo = JsonConvert.DeserializeObject<SoundCloudResponse>(json);
+                                        }
+                                    }
+                                }, TaskContinuationOptions.AttachedToParent);
+                    });
+
+            parentTask.RunSynchronously();
+
+            return new ContentProviderResultModel {Title = widgetInfo.Title, Content = widgetInfo.FrameMarkup};
```
davidfowl: Fix formatting.

JabbR/ContentProviders/SoundCloudContentProvider.cs
```diff
@@ -1,47 +1,64 @@
...
+                        var task = Task<WebResponse>.Factory.FromAsync(
+                            webRequest.BeginGetResponse, webRequest.EndGetResponse,
+                            TaskCreationOptions.AttachedToParent);
+
+                        task.ContinueWith(
+                            tr =>
+                                {
+                                    using (var stream = tr.Result.GetResponseStream())
+                                    {
+                                        if (stream == null)
```
davidfowl, line 34: No self closing ifs. Add braces.

alexkopy:
Measure? Original fetches 381495 bytes (almost none of which are needed) vs new fetching 921 bytes (almost all of which are needed). And new doesn't need to grep anything. And doesn't rely on webpage layout not changing (which is less trustworthy than an api special build for the purpose).

PR #381, opened 2012-02-15
fix a bug where emoji appears on pastie provider

reyrahadian:


davidfowl:
Can't fix this for just pasite. You need to do it for all of the providers that bring in code.

reyrahadian:
hmm.. correct me if i'm wrong, but this code should be able to handle all provider that extend CollapsibleContentProvider,  because the provider would always contain [<div class=""collapsible_box"">]. And we don't need to handle provider that extend EmbedContentProvider because they render embed tag

davidfowl:
Ok verify that it works for all providers that bring in code.

PR #382, opened 2012-02-16
Modify content providers so that they dont request pasted URLs unless they really need to

paulduran:
I'm using these providers in my campfire web client. I realise its slightly different in JabbR because they are requested only once but in my client, the URLs are requested whenever somebody new joins the room or refreshes their page etc.

It seems very inefficient and a possible risk to blindly request every URL that is pasted in any channel.

I was speaking to dfowler about it the other night and he mentioned they might be being re-written to be client-side however this might be a useful interim measure.

### DIFF
--- a/JabbR/ContentProviders/Core/ResourceProcessor.cs
+++ b/JabbR/ContentProviders/Core/ResourceProcessor.cs
@@ -10,19 +10,16 @@ namespace JabbR.ContentProviders.Core
     public class ResourceProcessor : IResourceProcessor
     {
         private readonly Lazy<IList<IContentProvider>> _contentProviders = new Lazy<IList<IContentProvider>>(GetContentProviders);
-        private const string _userAgent = "Mozilla/5.0 (compatible; MSIE 9.0; Windows NT 6.1; WOW64; Trident/5.0; MAAU)";
 
         public Task<ContentProviderResultModel> ExtractResource(string url)
         {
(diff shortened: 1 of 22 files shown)


JabbR/ContentProviders/Core/ResourceProcessor.cs
```diff
@@ -10,19 +10,16 @@ namespace JabbR.ContentProviders.Core
...
-        private const string _userAgent = "Mozilla/5.0 (compatible; MSIE 9.0; Windows NT 6.1; WOW64; Trident/5.0; MAAU)";
 
         public Task<ContentProviderResultModel> ExtractResource(string url)
         {
-            var request = (HttpWebRequest)HttpWebRequest.Create(url);
-            request.UserAgent = _userAgent;
-            var requestTask = Task.Factory.FromAsync((cb, state) => request.BeginGetResponse(cb, state), ar => request.EndGetResponse(ar), null);
-            return requestTask.ContinueWith(task => ExtractContent((HttpWebResponse)task.Result));
+            var uri = new Uri(url);
+            return Task.Factory.StartNew(() => ExtractContent(uri));
```
davidfowl, line 17: This looks wrong. I'd make ExtractContent return Task<ContentProviderResultModel> in case the content provider did want to make an http request.

davidfowl:
Any updates on this?

PR #385, opened 2012-02-20
Command refactor

graemechristie:
Refactored the Command functionality from a monolothic CommandManager class into a light command factory with individual commands in concrete classes. Command factory utilises Ninject to fetch commands (which are registered with the IOC container, currently using reflection, this could easily be refactored to static list of types in the factory class for perf reasons).

Commands can now be extended without touching existing code. Theoretically should be more in line with SRP and more testable. I'm not 100% the one class per command model is the way to go, and that the IOC container is the right place for individual commands to be fetched from. Seeking advise from the team on this.

Note this Pull request is just for review and should not be merged. Most tests still need to be refactored.


davidfowl:
1. Can we use MEF for the commands?
2. Can you make those changes within the CommandManager so none of the tests have to change.

The general idea looks good. I'll go into more details when you finish the changes.

graemechristie:
1. We could use MEF for the Commands...however we would still need to instantiate them using Ninject to resolve the dependancies (IRepository, IChatService, ISomeThingElseInTheFuture)  or use MEF to resolve the dependencies (and register them all with MEF, which is a bit redundant). I think just avoiding reflection and wiring up the commands manually (by looping though and array of concrete types in the bootstrapper) might alleviate some of the smell around this. I'm pretty sure registering the commands with Ninject in the bootstrapper is the right thing to do ... If we feel there are too many objects flying around, we register Command "Modules" with the Container, which each handle command as methods... same idea, less objects .. also would give us the ablilty to easily share functionality between commands with private methods).
2. I can re-introduce the command manager ... however I'll still need to change the test a bit as command manager doesn't require the various services in it's constructor and we shouldn't really leave them there. Manually new-ing up a a command manager within the Notification Service/Hub seems bad. 

---

Question - Is there a way we can get the instance of the INotificationService into the IOC Container. Currently the Instance we want is there as a Hub (the Hub instance implements INotificationService,) I don't know enough about the lifecycle of the Hub to suggest the best way of registering an instance as both the IHub and the INotificationService. This would save me from manually injecting it and passing it from command manager to command etc

davidfowl:
1. To I'd rather you make as little changes as possible so you don't break things.

To your last question no, just leave that par the way it is, its not injectable (plus there's circular reference)

graemechristie:
No circular reference if you use a Factory or Func<T>/Lazy<T> for the ICommand. I'll have a bit more of a think about things.

graemechristie:
OK. I've done a quick re-write using MEF, and keeping the interface for CommandManager unchanged. Tests pass as written. Some parts are a bit ugly e.g. Wiring the Notification service instances up to MEF using ComposeExportedValue<T> and instantiating a CommandManager with dummy args just to get at the command metadata for help ... It's late now, I'll look at these with new eyes tomorrow.

JabbR/.gitignore
```diff
@@ -0,0 +1 @@
+*.config
```
davidfowl, line 1: Revert.

JabbR/App_Start/Bootstrapper.cs
```diff
@@ -21,9 +21,12 @@
 using SignalR.Hosting.AspNet;
 using SignalR.Infrastructure;
 using SignalR.Ninject;
+using JabbR.Commands;
+using Ninject.Parameters;
+using SignalR.Hubs;
+using System.Reflection;
```
davidfowl, line 27: Sort usings.

JabbR/Auth/Login.ashx.cs
```diff
@@ -18,7 +18,7 @@ public class Login : IHttpHandler
         private const string VerifyTokenUrl = "https://rpxnow.com/api/v2/auth_info?apiKey={0}&token={1}";
 
         public void ProcessRequest(HttpContext context)
-        {
+       {
```
davidfowl, line 21: Fix formatting.

JabbR/App_Start/Bootstrapper.cs
```diff
@@ -68,9 +71,9 @@ public static void PreAppStart()
                 .InSingletonScope();
 
             kernel.Bind<IApplicationSettings>()
-                  .To<ApplicationSettings>()
-                  .InSingletonScope();
-
```
davidfowl, line 73: Fix formatting

JabbR/Commands/AddOwnerCommand.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Web;
+using JabbR.Services;
+using JabbR.Models;
+using Ninject;
+using System.ComponentModel.Composition;
```
davidfowl, line 8: Sort usings.

JabbR/Commands/AddOwnerCommand.cs
```diff
@@ -0,0 +1,58 @@
...
+
+namespace JabbR.Commands
+{
+    [Export(typeof(ICommand))]
+    [CommandMetadata(Name = "addowner", Usage = "Type /addowner [user] [room] - To add an owner a user as an owner to the specified room. Only works if you're an owner of that room.", Weight = 16.0f)]
```
davidfowl, line 13: Split long ones into multiple lines.

JabbR/Commands/AddOwnerCommand.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Linq;
+using System.Web;
+using JabbR.Services;
+using JabbR.Models;
+using Ninject;
+using System.ComponentModel.Composition;
+
+namespace JabbR.Commands
+{
+    [Export(typeof(ICommand))]
```
davidfowl, line 12: Instead of Exporting each command, add [InheritedExport] to ICommand.

JabbR/Commands/CommandManager.cs
```diff
@@ -42,25 +48,26 @@ public class CommandManager
...
+
+        private void LoadCommands() 
         {
-            commandName = commandName.Trim();
-            if (commandName.StartsWith("/"))
-            {
-                return false;
-            }
 
-            if (!TryHandleBaseCommand(commandName, parts) &&
```
davidfowl, line 55: Is this ordering preserved?

JabbR/Commands/CommandManager.cs
```diff
@@ -42,25 +48,26 @@ public class CommandManager
...
-            return true;
+            // Add our injected dependancies to MEF's Compoisition Container. 
+            // Bit of a hack as we are kind of bridging Ninject and MEF
+            container.ComposeExportedValue<INotificationService>(_notificationService);
+            container.ComposeExportedValue<IJabbrRepository>(_repository);
+            container.ComposeExportedValue<IChatService>(_chatService);
+
+            var batch = new CompositionBatch();
+            batch.AddPart(this);
+            container.Compose(batch);
```
davidfowl, line 70: Not sure about this part. Need to think about this code some more.

JabbR/Commands/CommandManager.cs
```diff
@@ -77,838 +84,32 @@ public bool TryHandleCommand(string command)
...
-                _repository.CommitChanges();
-
-                _notificationService.NudgeRoom(room, user);
+                command = _commands.First(lc => lc.Metadata.Name == commandName).Value;
             }
-            else
+            catch (InvalidOperationException) // Command was not found in the collection
             {
-                throw new InvalidOperationException(String.Format("Room can only be nudged once every {0} seconds", betweenNudges.TotalSeconds));
+                throw new InvalidOperationException(String.Format("'{0}' is not a valid command.", commandName));
```
davidfowl, line 102: Why not just check for null and throw instead of throwing twice?

JabbR/Hubs/Chat.cs
```diff
@@ -222,35 +223,15 @@ public Task Disconnect()
...
-                new { Name = "topic", Description = "Type /topic [topic] to set the room topic. Type /topic to clear the room's topic." }
-            };
+            // Map the Name & Description properties from all the registered command objects to an array of anonymous objects
+
+            // TODO: Ugly ... refactor somehow to get rid of this ugly not_service->command manager->help command->notification service->command_manager loop
+            var commandManager = new CommandManager("clientId", "userId", UserAgent, "roomName", _service, _repository, this);
+
+            var commandMetaData = commandManager.GetAllCommandMetaData();
+
+            return commandMetaData.Count() > 0 ? commandMetaData.Select(c => new { Name = c.Name, Description = c.Usage }).ToArray() : new object { };
```
davidfowl, line 233: Why not make this static? Commands don't change dynamically.

JabbR/index.htm
```diff
@@ -63,9 +63,9 @@
             e.id = 'janrainAuthWidget';
 
             if (document.location.protocol === 'https:') {
-                e.src = 'https://rpxnow.com/js/lib/jabbr/engage.js';
+                e.src = 'https://rpxnow.com/js/lib/gamr/engage.js';
```
davidfowl, line 66: This shouldn't have been committed.

JabbR/packages.config
```diff
@@ -1,31 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
davidfowl, line 1: What?

JabbR/Web.config
```diff
@@ -1,62 +0,0 @@
```
davidfowl: What?

JabbR/Web.Release.config
```diff
@@ -1,31 +0,0 @@
```
davidfowl: What?

JabbR/Web.Debug.config
```diff
@@ -1,30 +0,0 @@
-﻿<?xml version="1.0"?>
```
davidfowl, line 1: What?

JabbR/Commands/AddOwnerCommand.cs
graemechristie, line 12: Turns out [InheritedExport] is a bit broken when used with Metadata attributes.

http://stackoverflow.com/questions/6570930/how-to-use-mef-inherited-export-metadata

Calling the base constructor with typeof(ICommand) in CommandAttribute achieves much the same thing, so I have done that.

PR #392, opened 2012-02-24
This will allow url's to contain ()

admsteck:
Changed the urlPattern regex to allow url's to contain a matching pair of parenthesis.  
Other changes include:
- Added mailto: addresses
- Previous code was case sensitive and would find http:// but not HTTP://


admsteck:
This seemed like an easier one for my first experience with github and open source contribution.  Let me know if I messed anything up.

davidfowl:
Do all the tests pass? Can you add some more tests to verify your changes.

Issue #336, opened 2012-01-29
Can't scroll to see previous chat history on high resolution monitors


TrentScholl:
At 2560x1440, not enough previous chat messages are loaded to cause the scroll bar to show (unless images/youtube clips are in the history). This means you can't scroll up and trigger it to load previous history.

Workaround: Shrink the window down, scroll up and then maximize the window again.

Suggested Fixes: On room enter, display more messages regardless of resolution or on room enter, display number of messages based on resolution.

davidfowl:
Yes this was changed to make the loading faster but we can see what number of messages works best.

davidfowl:
Was fixed

PR #415, opened 2012-03-12
Added check for no scrollbar when joining a room and load more messages ...

admsteck:
...if no scrollbar present

For issue #336

### DIFF
--- a/JabbR/Chat.js
+++ b/JabbR/Chat.js
@@ -71,6 +71,11 @@
                         ui.addChatMessage(viewModel, room);
                     });
 
+                    // load more messages if no scrollbar is visible
+                    if (!ui.hasScrollbar(room) && roomInfo.RecentMessages.length > 0) {
+                        $ui.trigger(ui.events.scrollRoomTop, [{ name: room, messageId: roomInfo.RecentMessages[0].Id}]);
+                    };
+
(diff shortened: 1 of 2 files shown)


samandmoore:
@admsteck so I tested this a bit. I'm just going to throw in some test cases that I was using.

**NOTE:** I do not have a high resolution monitor, so I changed the initial message load to 5 messages in Chat.cs "GetRoomInfo" in order to simulate "no scrollbar"

This fix works when you change rooms. It will load more messages when the event `activeRoomChanged` is fired, but only if `messageId` is populated. In other words, when the "join" command is issued or when the initial load of the page happens, this fix does not work. 

### Test case 1: load directly to a room /#/rooms/roomname

_Expected Result:_ loads 5 messages, fires event to load more messages because there is more room and no scrollbar
_Actual Result:_ only the initial 5 messages are loaded.
_Suspected Issue:_ the `messageId` value is not set during `setActiveRoom` (perhaps because the messages have not been populated yet).

### Test case 2: issue command, /join roomname

_Expected Result:_ loads 5 messages, fires event to load more messages because there is more room and no scrollbar
_Actual Result:_ only the initial 5 messages are loaded.
_Suspected Issue:_ the `messageId` value is not set during `setActiveRoom` (perhaps because the messages have not been populated yet).

Issue #411, opened 2012-03-12
Typing " followed by ) causes a wink smiley


davidfowl:
It's because of html encoding. So " => `&quot;)`

davidfowl:
Fixed

PR #418, opened 2012-03-13
fix for issue #411 - added encode/decode html functions to prevent emoji from parsing encoded html entities

samandmoore:
I changed the emoji parser to use two new utility functions: encodeHtml and decodeHtml. First it fully encodes the message content then it decodes it and parses it for emoji. This fixes the issue where ") encodes as &quot;) which displays the wink emoji.


JabbR/Chat.utility.js
```diff
@@ -102,8 +114,11 @@
         isMobile: isMobile,
         parseEmojis: function (content) {
             var parser = new Emoji.Parser().parse;
-            return (parser(content));
+            var decodedContent = encodeHtml(decodeHtml(content));
```
davidfowl: This has been tried before and it's breaks urls.

samandmoore: hmm, I see that now... My other idea is to force emoji to be prefixed with a space so that "some text;)" would not display winkface, but "some text ;)" would display winkface. that wouldn't require any fiddling with html encoding/decoding.

EDIT: I'd have to come up with a solution that would work for just ";)" though since it is valid and doesn't have whitespace prefix... hmm i will have to give this more thought.

samandmoore:
@davidfowl I decided to switch up my strategy. I'm now handling the issue in the Chat.emoji.js file instead of in the Chat.utility.js file. This fixes what you pointed out with the links no longer working. The idea is that it wraps the submitted content in html then only searches for emoji inside the innerText of html elements. It looks like it fixed the link-breakage, it still avoids any script injection attempts, and it properly parses the emoji.

davidfowl:
Awesome, let's see if anything else breaks. You can test this on staging => jabbr-staging.apphb.com

samandmoore:
looks like things are working on staging. I've tried a bunch of content providers too.

davidfowl:
What exactly did you fix? I'm still seeing this error ") causes the ;) smiley.

samandmoore:
on jabbr-staging.apphb.com? I'm seeing it works fine, typing ") displays ") not ;) ... Maybe your browser is caching the js files?

davidfowl:
yay works now :smile:

samandmoore:
yay :smile:

PR #423, opened 2012-03-18
Replaced AddClient with UpdateActivity

larrybeall:
This both updates the user activity and adds the client fixing the issue.


larrybeall:
It appears that this also fixes another issue I'm having.  When you log in with /nick and auto-join the room the user does not show in the list.  After this code change that issue is corrected.

Issue #428, opened 2012-03-29
Downloaded messages aren't sorted in chronological order


thecodejunkie:
https://github.com/davidfowl/JabbR/blob/master/JabbR/api/MessagesHandler.cs#L88

davidfowl:
Fixed

PR #431, opened 2012-04-04
Adding OrderBy for Issue #428

samandmoore:
Added OrderBy using the message date (When) to ensure that sorting is by message's datetimeoffset value when you download the logs for a room.


davidfowl:
This pull request has 11 commits. Something went wrong here. Start over please :)

Issue #424, opened 2012-03-20
Can not login with Stack Exchange OpenID


MoriTanosuke:
When I try to log in with my own domain and _Stack Exchange Open ID_ as delegate, I get this error message:

```
[InvalidOperationException: 'Carsten Ringe' is not a valid user name.]
   JabbR.Services.ChatService.AddUser(String userName, String identity, String email) in c:\dev\github\JabbR\JabbR\Services\ChatService.cs:287
   JabbR.Auth.Login.ProcessRequest(HttpContext context) in c:\dev\github\JabbR\JabbR\Auth\Login.ashx.cs:93
   System.Web.CallHandlerExecutionStep.System.Web.HttpApplication.IExecutionStep.Execute() +625
   System.Web.HttpApplication.ExecuteStep(IExecutionStep step, Boolean& completedSynchronously) +270
```

davidfowl:
It's because of spaces in your name. We could autofix up these types of logins.

samandmoore:
This was fixed in PR #432

Issue #395, opened 2012-02-28
Unhandled Exception (YSOD) logging in via OpenID


factormystic:
_(Thank you for supporting OpenID!)_

Guessing it might be the whitespace in my username, but the system should be resilient to this input without a ysod

```
Exception Details: System.InvalidOperationException: 'Factor Mystic' is not a valid user name.

Source Error: 

An unhandled exception was generated during the execution of the current web request. Information regarding the origin and location of the exception can be identified using the exception stack trace below.

Stack Trace: 


[InvalidOperationException: 'Factor Mystic' is not a valid user name.]
   JabbR.Services.ChatService.AddUser(String userName, String identity, String email) in C:\dev\github\JabbR\JabbR\Services\ChatService.cs:287
   JabbR.Auth.Login.ProcessRequest(HttpContext context) in C:\dev\github\JabbR\JabbR\Auth\Login.ashx.cs:93
   System.Web.CallHandlerExecutionStep.System.Web.HttpApplication.IExecutionStep.Execute() +625
   System.Web.HttpApplication.ExecuteStep(IExecutionStep step, Boolean& completedSynchronously) +270
```

davidfowl:
Ah good one. We maybe need to fix up your name.

samandmoore:
this was fixed in PR #432 :)

PR #432, opened 2012-04-04
fixing issue #428 and #424

samandmoore:
#428 - sorting the messages by "When" ascending when downloading
#424 - replacing spaces with underscores (may want to increase the amount of chars we replace pending OpenID spec review)

### DIFF
--- a/JabbR/Auth/Login.ashx.cs
+++ b/JabbR/Auth/Login.ashx.cs
@@ -117,7 +117,7 @@ namespace JabbR.Auth
 
         private string FixUserName(string username) {
             // simple for now, translate spaces to underscores
-            return username = username.Replace(' ', '_');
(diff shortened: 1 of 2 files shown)


JabbR/Auth/Login.ashx.cs
```diff
@@ -113,6 +115,11 @@ public void ProcessRequest(HttpContext context)
             context.ApplicationInstance.CompleteRequest();
         }
 
+        private string CleanUpOpenIDUsername(string username) {
```
davidfowl: Change the name to FixUserName. This isn't specific to OpenID (hopefully).

samandmoore:
this PR also fixed this issue #395 https://github.com/davidfowl/JabbR/issues/395 which was a dupe of #424

PR #434, opened 2012-04-05
Added support for 9gag as content provider

mikehc:
What the tittle says. Csproj really didn't changed that much, only for the addition of the ContentProvider, didn't know what happened. Totally n00b here.


davidfowl:
I see tabs and did you set auto crlf to true? Check out the readme and send another pull request. Also I'm not sure why the csproj has so many changes but I don't trust it.

davidfowl:
Send a new one with those fixes.

mikehc:
Ok, I'll check those things and send another one.

Thank you.

PR #436, opened 2012-04-07
adding jquery.keybind plug-in for shift+tab room navi

samandmoore:
this adds the ability to cycle through tabs via shift+tab keybinding


davidfowl:
What about the access keys.

samandmoore:
I wasn't sure how to trigger the access keys. But if that's the preferred
method we can dump this PR. I just prefer quick cycling of tabs.

PR #441, opened 2012-04-12
Modified Jabbr to use a SQL Server 2012 Express LocalDB database by defa...

DanRigby:
...ult (no SQL Server service required).


davidfowl:
Not yet. Since I don't have it.

PR #443, opened 2012-04-12
Added /invite command.

DanRigby:


davidfowl:
No description? Just felt like adding a new feature?

DanRigby:
It's 5am. :D So the idea is just that provides a slightly more intuitive syntax for easily inviting people to join another room.

davidfowl:
You can verify it on staging http://jabbr-staging.apphb.com

PR #451, opened 2012-04-20
Adding fk to models

samandmoore:
- I moved all the EF configuration code into classes using the EF conventions.
- I added foreign key properties to all of the models that were missing them


JabbR/Models/JabbrContext.cs
```diff
@@ -7,33 +8,18 @@ public class JabbrContext : DbContext
         public JabbrContext()
             : base("Jabbr")
         {
+            Database.SetInitializer<JabbrContext>(null);
```
davidfowl: Why do you need this?

samandmoore: I'm just being explicit.

According to http://msdn.microsoft.com/en-us/library/gg679461(v=VS.103).aspx the default is: 

> "The default strategy is an instance of DropCreateDatabaseIfModelChanges<TContext> created with useSeedData set to true."

And I figure that we don't want that, so I'm just explicitly saying, "We don't want to use any DbInitializer, kthxbai"

davidfowl: Don't need that since we use Migrations.

samandmoore: okay, i will remove it.

davidfowl:
Ok looks good.

JabbR/Models/Mapping/ChatRoomMap.cs
```diff
@@ -0,0 +1,66 @@
...
+            this.HasMany(r => r.Users)
+                .WithMany(u => u.Rooms)
+                .Map(m =>
+                    {
+                        m.ToTable("ChatUserChatRooms");
+                        m.MapLeftKey("ChatRoom_Key");
+                        m.MapRightKey("ChatUser_Key");
+                    });
+
+            this.HasOptional(r => r.Creator)
```
davidfowl, line 60: This is wrong.

samandmoore, line 60: How so? I don't see the bug.

davidfowl, line 60: This is a 1:1 not 1:many

samandmoore, line 60: which relationship are you talking about? As far as I can tell these are exactly the same as in the old OnModelCreating method, no?
- A `Room` has many `User`s that are allowed to be in it
- A `User` has many `Room`s they are allowed to be in
- A `Room` has many owners
- A `User` can be owner in many `Room`s
- A `User` can be chatting in many `Room`s
- A `Room` can have many `User`s chatting in it
- A `Room` may have a `User` that created it
- A `User` can be creator in many `Room`s

PR #453, opened 2012-04-23
Upgrade EF 4.3.1

samandmoore:
updated all of the migrations to the new API.

renamed the migrations file to follow with the new convention. 

added migration to remove EdmMetadata table.


JabbR/Migrations/201204232322474_RemoveEdmMetadata.cs
```diff
@@ -0,0 +1,25 @@
+namespace JabbR.Migrations
```
davidfowl: What's this?

samandmoore: VS hates me. changed.

PR #454, opened 2012-04-25
Upgrading to Entity Framework 4.3.1

samandmoore:
- removed EF Migrations package
- updated EF Core package
- updated all existing Migrations to use new Migrations API (required)
- renamed Migrations Settings.cs file to MigrationsConfiguration.cs to fall in line with API standards
- changed namespace of MigrationsConfiguration to be the same as Migrations files (ensures that new Migrations will default into the same namespace)
- Added a migration to remove the EdmMetadata table (required change for 4.3.1)... this migration also triggers an internal change, it changes the schema of the __MigrationHistory table (this is something EF does when going from 4.2 to 4.3.1).

To be safe, I'd recommend backing up the DB before pushing this live. I've verified that it works properly and migrates the changes properly on multiple test instances both locally and on AppHarbor, but "it's better to be safe than sorry."


JabbR/App_Start/Bootstrapper.cs
```diff
@@ -132,7 +131,7 @@ private static void DoMigrations()
             }
 
             // Only run migrations for SQL server (Sql ce not supported as yet)
-            var settings = new Settings();
+            var settings = new JabbR.Models.Migrations.MigrationsConfiguration();
```
davidfowl, line 134: Why full type name here?

samandmoore, line 134: There is a Migration and a Model with the same name `UserStatus` so including the `JabbR.Models.Migrations` as a using statement won't compile. Rather than trying to change the name of any files I opted to just fully qualify the type.

PR #468, opened 2012-05-13
Adding ARIA attributes and re-organize modals for main page readablity

nschonni:
Adding ARIA attributes to indicate application sections and tabs for screen reader devices.
Includes turning of static content caching for CSS and JS files.


davidfowl:
Does this have any effect on the layout?

nschonni:
No, it changes the underlying structure of the HTML with semantic elements but visually the rendering looks the same.

Issue #386, opened 2012-02-20
Disable notification on rooms


thecodejunkie:
It would be handy to disable notification on any room you'd like. Some rooms aren't that important to me and I'd like to avoid getting a seizure of I don't check the updates of the room ;)

ctolkien:
Further on this, I think it would be good to stop the blinking after X number of blinks

davidfowl:
@ctolkien's suggestion is a good balance.

thecodejunkie:
I disagree this doesn't really touch at the reported issue. If you have several tabs open and some you don't really care about notifications then you are still going to get epilepsy =) A busy room will still blink for as long as you keep the tab open

davidfowl:
Yep but blinking less eases the pain. I went for a compromise that is
think is reasonable.

Sent from my Windows Phone

drub0y:
@davidfowl Totally agree with this as compromise, but do you think you could have it end the blinking on the highlighted color? That way when we're not staring at the screen and come back we can distinguis what rooms actually had activity since last we looked?

davidfowl:
Sounds good.

Sent from my Windows Phone

PR #472, opened 2012-05-14

drub0y:
This PR is mainly about discussion around #386 and basically adds a lasting highlight to tabs that haven't yet been read. However there were also several files with incorrect line endings that I had to fix to get my repo normalized. JabbR README clearly states contributors must be using core.autocrlf=true and that's what I'm using, so someone else must not be.

I chose a purplish color for the highlight so it stands out from the dark blue background, but also doesn't slap you in the face like a brighter color might. Feel free to change that to something else that suits your design.


davidfowl:
:smile: Redo?

Issue #76, opened 2011-11-16
Feature @Username


coden4fun:
When @username is displayed mouse over and show small info's of people of their gravatar facebook or twitter 

This would be helpful so people wouldn't have to ask hey can i get your x account.

Simply just roll mouse over and wait one second and there are the links.

davidfowl:
Where would that information come from?

coden4fun:
well you'd have to build ways for me to use my facebook/twitter account but we can use gravatar right?  So I would assume it would come from the service you're proving us to login?  (i.e. add facebook and twitter and a way for me to access my facebook by knowing i'm always coden4fun in the room)

thedalewilliams:
Can you just use gravatar?

http://en.gravatar.com/ea0cd17bcc28810e1581fa9881298c60.json

ghost:
that would be a better option, but I think you're assuming everyone would have Gravatar by default before logging in to the chat.  What if the user doesn't have a Gravatar?

The json is nice as I can get your twitter, facebook, linkedin, etc.. but for those who don't use Gravatar?  Are you going to force them to have a Gravatar before they use the product?

davidfowl:
Yes, everyone should have a gravatar. Plus it saves us a bunch of work :)

Buildstarted:
Thanks for that link enslam. I have a partial mockup going for this. :)

http://i.imgur.com/Ds7XZ.png

davidfowl:
@Buildstarted What happened to this :smile:?

Buildstarted:
I lost my weekends, @davidfowl. I will resume work on this asap :)

PR #475, opened 2012-05-15
Added gravatar information to "/who" command. Start for issue #76

nschonni:
Also inlined Janrain script reference to match Google method so server-side script variable is called less time.


nschonni:
Closing because of unrelated change. Will submit a new PR in fixed branch.

PR #481, opened 2012-05-15

drub0y:
Addresses discussion in #386 where existing implementation will eventually go back to looking like a normal tab that never had any activity. This change will glow the tab to a highlight color on the final glow.


davidfowl:
You changed my braces :). Try again

drub0y:
Damn VS auto-format on paste! Sorry, one sec.

PR #482, opened 2012-05-15

drub0y:
Addresses discussion in #386 where existing implementation will eventually go back to looking like a normal tab that never had any activity. This change will glow the tab to a highlight color on the final glow.


drub0y:
Just noticed I said "Leave the table highlighted" in the comment (instead of "tab"), fixing that real quick.

PR #485, opened 2012-05-16
Message handler now works using Web API

izevaka:
The bitch has been Web APIed.


JabbR/Infrastructure/JsonNetMediaFormatter.cs
```diff
@@ -0,0 +1,85 @@
...
+using System.Net.Http.Headers;
+using System.Text;
+using System.IO;
+using Newtonsoft.Json.Serialization;
+using Newtonsoft.Json.Converters;
+using System.Net;
+
+namespace JabbR.Infrastructure
+{
+    public class JsonNetMediaFormatter : MediaTypeFormatter
```
davidfowl, line 17: You shouldn't need this type at all. JSON.NET is built in no?

davidfowl:
OK so here's wht I did, I took your code and updated to the latest nightly build and it works great. It's in master so you can discard this branch.

PR #490, opened 2012-05-18
Moved common API response code into extension methods.

izevaka:
- Added QueryString extension method that tokenizes URL querystring
  parameters.
- Added TryGetAndConvert method that gets a querystring parameter and tries
  to convert it to a specified type
- Created a whole bunch of extension methods to return success and error
  responses.


JabbR/api/MessagesController.cs
```diff
@@ -20,7 +21,7 @@ public MessagesController(IJabbrRepository repository)
             _repository = repository;
         }
 
-        public HttpResponseMessage GetAllMessages(string room, string range, bool download = false)
+        public HttpResponseMessage GetAllMessages(string room, string range)
         {
```
davidfowl, line 25: Where's the download parameter?

JabbR/Infrastructure/UriExtensions.cs
```diff
@@ -0,0 +1,96 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Web;
+using System.Web.Http.Routing;
+using System.ComponentModel;
+
+namespace JabbR.Infrastructure
+{
+    public static class UriExtensions
```
davidfowl: Why do you need this class?

izevaka: Because outside of action, parameter/model binding is not available. I would imagine that we'd need this when we need to parse api key (probably in an authorization filter. Also when adding jsonp support. Not sure where jsonp bit would sit, probably needs a customized mediaformatter with a bit of code in delegatingmessagehandler to set content-type.

JabbR/api/MessagesController.cs
izevaka, line 25: That's now in the common bit here - https://github.com/davidfowl/JabbR/pull/490/files#diff-3.

The thinking is that download parameter can be applied to any api call, so I wanted to move it out of any specific controllers, to keep them cleaner.

JabbR/Infrastructure/UriExtensions.cs
davidfowl: Can we add it when we need it? I don't like things being there for the sake of it just in case we're going to need it. Bring it back when it's being used.

JabbR/api/MessagesController.cs
davidfowl, line 25: Ok this looks ok.

JabbR/Infrastructure/UriExtensions.cs
izevaka: It's already used now by AddDownloadHeader. I think I know what to do to make it a bit cleaner. How about creating a querystring collection class, that way, all the semantics of getting a value and converting it that are very specific to url parameter binding are not available on a generic IDictionary extension.

JabbR.Tests/QueryStringCollectionTests.cs
```diff
@@ -0,0 +1,154 @@
...
+                });
+            }
+
+        }
+        public class TryGetAndConvert
+        {
+            QueryStringCollection _QueryStringCollection;
+            public TryGetAndConvert()
+            {
+                _QueryStringCollection = new QueryStringCollection(new Uri("http://example.com?intvalue=1&stringvalue=str&boolvalue=true&empty"));
```
davidfowl: I'm not a fan of class level arguments in unit tests, can you move this to each test? It makes it clearer as you don't have to look outside the test to get context.

davidfowl:
Rebased.

PR #494, opened 2012-05-19
Plumbing in support to be a portable web application

loudej:
Uses OWIN server abstractions


JabbR/Startup.cs
```diff
@@ -0,0 +1,184 @@
...
+                            app(
+                                env,
+                                (status, headers, body) => result(
+                                    status,
+                                    headers,
+                                    (write, flush, end, cancellationToken) => ExecutionContext.Run(
+                                        context.CreateCopy(),
+                                        __ => body(write, flush, end, cancellationToken),
+                                        null)),
+                                fault);
```
samandmoore: this is really difficult to read, is there a better way to format this?

loudej: I removed it altogether

that particular middleware was just being defensive - in case the http server didn't provide a new ExecutionContext per http connection. In the end it probably makes more sense to require the http server handlers for firefly and gate to ensure there's a new execution context that "sticks" to each of the calls coming in for a request.

loudej:
This will take some refactoring at other levels before a pull request to jabbr is practical.

Thanks for the feedback though!

PR #495, opened 2012-05-20
Animate topic changes

dpen2000:
This is promised topic change animation

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -690,13 +690,13 @@
         var topic = roomViewModel.Topic;
         var topicHtml = topic === '' ? 'You\'re chatting in ' + roomViewModel.Name : '<strong>Topic: </strong>' + topic;
         var roomTopic = room.roomTopic;
-        var isVisibleRoom = getCurrentRoomElements().getName() == roomViewModel.Name;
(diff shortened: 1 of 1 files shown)


JabbR/Chat.ui.js
```diff
@@ -689,7 +689,15 @@
...
         var topic = roomViewModel.Topic;
         var topicHtml = topic === '' ? 'You\'re chatting in ' + roomViewModel.Name : '<strong>Topic: </strong>' + topic;
-        room.roomTopic.html(topicHtml);
+        var roomTopic = room.roomTopic;
+        var isVisibleRoom = getCurrentRoomElements().getName() == roomViewModel.Name;
+        if (isVisibleRoom) {
+            roomTopic.hide();
+        }
+        roomTopic.html(topicHtml);
+        if (isVisibleRoom) {
```
samandmoore, line 698: is it necessary to check `isRoomVisible` twice?

dpen2000, line 698: See: https://github.com/davidfowl/JabbR/pull/495#commitcomment-1353163

samandmoore, line 698: we're not liking the `slideDown` it looks weird... maybe try `roomTopic.fadeIn(2000);`? that would be using the same fade effect that the rotating tips use that are below the message box. if you wanna change it to that we will take another look.

PR #498, opened 2012-05-23
First attempt at making JabbR responsive (using bootstrap)

AlecTaylor:
Alpha stage. Don't accept for the moment, but do report bugs :)


samandmoore:
I don't see any benefit from these changes.

PR #502, opened 2012-06-06
Added API entry point.

izevaka:
- Added ApiFrontPageController that serves up Janrain appId and api v2
  messages URL template.
- Renamed api directory to WebApi. This is to enable serving of /api URI.
  When a path is requested, IIS by default checks if a directory exists and
  would try to serve it, taking precedence over routes. This can be fixed by
  configuring IIS, but is a lot more work.
- Fixed up namespaces of all the files under WebApi directory to reflect the
  folder structure.


davidfowl:
Didnt' even see this :)

samandmoore:
does this need to be authenticated in some way? or is it okay to publicly expose the janrain app id?

izevaka:
Pretty sure app id is not a secret. In fact janrain console states that API key is secret and app id is not. I think they use it for configuration (ie configuring which providers you want to use) and analytics. 

I haven't been able to 100% verify this, so here is the support ticket I lodged with Janrain: https://support.janrain.com/requests/4311

izevaka:
The janrain ticket got closed, here is the stackoverflow version. http://stackoverflow.com/questions/10993341/should-janrain-appid-be-secret

PR #503, opened 2012-06-08
Make YouTube content provider work with https URIs

cbley:
Adds https domains for YouTube content providers


JabbR/Chat.css
```diff
@@ -456,7 +456,7 @@ li.room
   width: 100%;
   height: 20px;
   resize: none;
-  overflow: hidden;
+  overflow: auto;
```
davidfowl, line 459: What's this?

cbley, line 459: When you're typing in the new message box and it wraps to a new line, this allows you to see the line you're typing and not the first line.

However, it's not supposed to be a part of this pull request as it has nothing to do with the https youtube issue. I think I might have incorrectly created this pull request because it keeps adding my new commits instead of being just up to cc563c8

davidfowl, line 459: Can you redo both fixes in separate pull requests?

Issue #513, opened 2012-06-19
Add /welcome command for per-room welcome message


nschonni:
The command would display similar to the welcome message in the Lobby, but only fired when a user first joins the room.
This would have the same rules as the /topic command. 
1. Only be set by a room owner or admin
2. Cleared when no parameters given

PR #514, opened 2012-06-20
Add /welcome command for per-room welcome message

thomasjo:
_This pull request is tied to #513._

Not sure if this is good enough, but it seems to work. Given I'm not very familiar with the code base yet, I've very likely added some changes in the wrong areas — just let me know what to change, and I'll fix it.


samandmoore:
everything looks good so far, I haven't had the chance to pull and verify yet though.

one issue is that this is just the command part of #513. this does not include this part:

"The command would display similar to the welcome message in the Lobby, **but only fired when a user first joins the room.**"

Can we add that?

nschonni:
You should also add the new command to the instructions dictionary https://github.com/thomasjo/jabbr/blob/92d72b8f2b098c208c462b0eaddda72573a416f6/JabbR/Hubs/Chat.cs#L236

davidfowl:
Where are the tests?

PR #520, opened 2012-06-27
See what you're typing when new message line wraps

cbley:
When the new message line wraps, you can no longer see what you're typing. This allows you to see the line you're typing on.

Tested in IE9 on Windows and the latest stables of Chrome, Firefox, Safari on Windows 7 and OS X

### DIFF
--- a/JabbR/Chat.css
+++ b/JabbR/Chat.css
@@ -456,7 +456,7 @@ li.room
   width: 100%;
   height: 20px;
   resize: none;
-  overflow: hidden;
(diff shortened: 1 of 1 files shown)


davidfowl:
Send a image for this one.

cbley:
This is an issue in Firefox. It's hard to see in screenshot, but the cursor is below the line you see. You can't see what you're typing after it wraps.

Before

Setting overflow: auto; fixes this, and Firefox will behave like the other browsers.

After

PR #521, opened 2012-06-27
Created a content provider for spotify links

thematlin:


JabbR/ContentProviders/SpotifyContentProvider.cs
```diff
@@ -0,0 +1,30 @@
...
+    public class SpotifyContentProvider : CollapsibleContentProvider
+    {
+        protected override Task<ContentProviderResult> GetCollapsibleContent(ContentProviderHttpRequest request)
+        {
+            var spotifyUri = ExtractSpotifyUri(request.RequestUri.AbsolutePath);
+
+            return TaskAsyncHelper.FromResult(new ContentProviderResult()
+                                                  {
+                                                      Content = string.Format("<iframe src=\"https://embed.spotify.com/?uri=spotify:{0}\" width=\"300\" height=\"380\" frameborder=\"0\" allowtransparency=\"true\"></iframe>", spotifyUri),
+                                                      Title = string.Format("spotify:track:{0}", spotifyUri)
```
davidfowl: Uppercase S in string.Format

JabbR/ContentProviders/SpotifyContentProvider.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        private string ExtractSpotifyUri(string requestUrl)
+        {
+            return requestUrl.Remove(0, 1).Replace('/', ':');
+        }
+
+        public override bool IsValidContent(Uri uri)
+        {
+            return uri.AbsoluteUri.StartsWith("http://open.spotify.com/", StringComparison.CurrentCultureIgnoreCase);
+        }
```
samandmoore, line 28: is there a https url as well? we should also check for that if there is one; in fact, we'd prefer it.

davidfowl, line 28: Doesn't matter if the link itself is https or not, it's more about the embedded content

samandmoore, line 28: Oh duh! Right. My bad I was reading the wrong line. But it would be good to
match an https URL as well :)
On Jun 27, 2012 4:55 PM, "David Fowler" <
reply@reply.github.com>
wrote:

thematlin, line 28: I tried using https on the spotify open link and didn't get any response so I assume they don't support it. Which is why I only match on http.

JabbR/ContentProviders/SpotifyContentProvider.cs
thematlin: Fixed in latest commit

PR #525, opened 2012-07-11
Add support for enterprise single sign on using WIF

woloski:
Implemented 
- WS-Federation + SAML tokens support
- WIF Configuration in AppSettings instead of default config section to simplify cloud deploy
- Identity provider selector if using Windows Azure Active Directory

demo here: http://jabbr-auth10.apphb.com

TODO
- refactor user - identity association of JanRain and FedAuth into a single class
- federated logout when doing /logout
- get rid of FedAuth cookie since Jabbr uses its own cookie by removing SessionAuthModule, need some testing


JabbR/Web.config
```diff
@@ -49,6 +60,11 @@
       <add verb="POST,GET,HEAD" path="elmah.axd" type="Elmah.ErrorLogPageFactory, Elmah" />
       <add path="*.less" verb="GET" type="dotless.Core.LessCssHttpHandler, dotless.Core" />
     </httpHandlers>
+    <authentication mode="None" />
+    <!-- comment this if you want to show the identity selctor (fedauth.waad.selectorEnabled = true)-->
+    <authorization>
+      <deny users="?" />
```
davidfowl, line 74: What? Why do you need this.

JabbR/default.aspx
```diff
@@ -62,6 +70,42 @@
...
+        })();
+
+        function waad_load(data) {
+            if (data.length == 0) {
+                $("#waadEmbed").append('<p>No identity providers were configured</p>');
+                return;
+            }
+
+            var list = '';
+            $.each(data, function (k, idp) {
```
davidfowl: k? better variable names please. Also move the identity provider code into it's own javascript file, add possibly use the templating to generate the html (the rest of jabbr uses templating).

JabbR/Web.config
```diff
@@ -39,7 +50,7 @@
     <pages validateRequest="false" />
     <customErrors mode="Off" />
     <compilation debug="true" targetFramework="4.0" />
-    <httpRuntime requestValidationMode="2.0" />
+    <httpRuntime requestValidationMode="2.0" requestValidationType="JabbR.App_Start.FederatedLogin+AllowTokenPostRequestValidator" />
```
davidfowl: Is this required? Won't it always run even when WIF isn't in the pipeline?

JabbR/App_Start/FederatedLogin.cs
```diff
@@ -0,0 +1,149 @@
...
+
+[assembly: WebActivator.PostApplicationStartMethod(typeof(JabbR.App_Start.FederatedLogin), "PostAppStart")]
+[assembly: WebActivator.PreApplicationStartMethod(typeof(JabbR.App_Start.FederatedLogin), "PreAppStart")]
+
+namespace JabbR.App_Start
+{
+    public static partial class FederatedLogin
+    {
+        public static string ApplicationRealm
+        {
```
davidfowl: Add _all_ config settings to `IApplicationSettings` and use this interface as the primary way to get setting information. Keep it DRY.

JabbR/App_Start/FederatedLogin.cs
```diff
@@ -0,0 +1,149 @@
...
+                    {
+                        return x509.Certificate.Subject;
+                    }
+                }
+
+                return null;
+            }
+        }
+
+        public class AllowTokenPostRequestValidator : RequestValidator
```
davidfowl: Move this to its own class.

JabbR/Content/IdentityProviderSelector.css
```diff
@@ -0,0 +1,6 @@
+﻿#waadEmbed { margin:0px auto; width: 400px; }
```
davidfowl: Format this css properly.

JabbR/Hubs/Chat.cs
```diff
@@ -12,6 +12,7 @@
 using JabbR.ViewModels;
 using Newtonsoft.Json;
 using SignalR.Hubs;
+using System.Configuration;
```
davidfowl: Sort usings.

brandwe:
After Thursday I will help out with this

JabbR/Web.config
woloski, line 74: WIF has an http module that will trigger the signin process if the PassiveRedirectEnabled flag is = true and the user is not authenticated (principal.Identity.IsAuthenticated). In general is a good practice to deny all anonymous users to a site. 

Jabbr is a single page app and for the case that you use the selector I override the flag PassiveRedirectEnabled and remove the authorization section so that the default.aspx page can be shown and the identity selector is rendered.

I could get rid of this, fallbacking to "jabbr.state" cookie as the mechanism to detect if the user is authenticated or not and redirect via javascript in default.aspx. This is also related to the last item in the TODO list... I think I will get rid of the WIF cookie all the way and just use WIF for the initial authentication. I will do the rest of the things first and then try to tackle this one which is more heavy.

JabbR/Web.config
woloski: Yes, this will always run... The validator will check if the POST is a WS-Federation response. I could add a check in the validator to see if "IdentityProviderUrl" was set. I will have to use the static Kernel to get IApplicationSettings though, something you don't like :)

JabbR/App_Start/FederatedLogin.cs
woloski: Ok, will have to use Kernel here as well. Let's see if I don't get into an issue of what Pre/Post runs first (Bootstrapper or this)

JabbR/Hubs/Chat.cs
woloski: Yes, I got into an issue because Kernel is not initialized when I access the configuration here. There is no way to specify order in which Pre/Post methods run http://stackoverflow.com/questions/9041558/define-an-initialization-order-of-webactivator-preapplicationstartmethod-classes?rq=1

woloski: Ok, we solved it together with @samandmoore on jabbr... basically we had to change Bootstrapper to run on PreAppStart (was running on Post) and add the Order attribute (first Bootstrapper, then FedLogin)

woloski:
``` cs
[assembly: WebActivator.PreApplicationStartMethod(typeof(JabbR.App_Start.Bootstrapper), "PreAppStart", Order = 0)]
[assembly: WebActivator.PreApplicationStartMethod(typeof(JabbR.App_Start.FederatedLogin), "PreAppStart", Order = 1)]
```

davidfowl: Change it back, it's that way for a reason.

woloski:
Ok, @davidfowl check out the latest code and let me know.

JabbR/Web.config
```diff
@@ -39,7 +49,7 @@
     <pages validateRequest="false" />
     <customErrors mode="Off" />
     <compilation debug="true" targetFramework="4.0" />
-    <httpRuntime requestValidationMode="2.0" />
+    <httpRuntime requestValidationMode="2.0" requestValidationType="JabbR.Auth.AllowTokenPostRequestValidator" />
```
davidfowl: Why does this need to be here and doesn't it run for every request?

JabbR/Services/ApplicationSettings.cs
```diff
@@ -36,5 +36,48 @@ public string AuthAppId
...
+                return ConfigurationManager.AppSettings["fedauth.certThumbprint"];
+            }
+        }
+
+        public bool FedAuthRequiresSsl
+        {
+            get
+            {
+                string requireSsl = ConfigurationManager.AppSettings["fedauth.requireSsl"];
+                return (!string.IsNullOrEmpty(requireSsl) && bool.Parse(requireSsl));
```
davidfowl: uppercase `String` and `Boolean`. And use TryParse

JabbR/Web.config
woloski: Yes it runs for every request. It's shortcircuited if fed auth is not enabled. 

https://github.com/davidfowl/JabbR/pull/525/files#L3L-1

JabbR/default.aspx
woloski: refactored on a5a867c

JabbR/App_Start/FederatedLogin.cs
woloski: refactored on 76649a0

JabbR/App_Start/FederatedLogin.cs
woloski: This was fixed on 91de801
The decision was to split the Bootstrapper in two phases
- PreAppStart: container initialization
- PostAppStart: setup routes, hubs, etc 

In that way other initializers (like FederatedLogin can use the container). See 14e8cd8 as well

JabbR/Hubs/Chat.cs
woloski: We finally decided to split it. See 91de801

The issue was that we had to use the container to access configuration when the modules were dynamically registered in FederatedLogin on PreAppStart. Apparently when using `DynamicModuleUtility.RegisterModule` HttpModule.Init is being called in the context of PreAppStart and since we had the Bootstrapper setting up the container on PostAppStart it failed with `ArgumentNullException`

JabbR/Services/ApplicationSettings.cs
woloski: Done. Fixed on 6f32e34

JabbR/Content/IdentityProviderSelector.css
woloski: Fixed on b6ff247

JabbR/Web.config
davidfowl: So you replace the default validator with one specific for federated login? Does it just call the default one if it fails?

woloski: Yes, that's what WIF recommends. The logic is not "call the default one if it fails", it's more like "always call the default one unless fed auth is enabled and the incoming request is wsfederation token post". 

This is the pseudocode of the validator:

```
if (!fedauthenabled)
   run default validator (base.validate())

if (request is wsfederation response)
   return request is valid

run default validator (base.validate())
```

JabbR/App_Start/FederatedLogin.Hooks.cs
```diff
@@ -0,0 +1,148 @@
...
+
+                if (identity.Name == null)
+                {
+                    throw new InvalidOperationException("Name claim not found.");
+                }
+
+                string userIdentity = userIdentityClaim.Value;
+                string username = identity.Name;
+                string email = String.Empty;
+                if (identity.Claims.SingleOrDefault(c => c.ClaimType == ClaimTypes.Email) != null)
```
davidfowl: Why do you run this query twice in a row if it's non null?

woloski: I can assign it to a variable if that's what you want. The perf cost is null since it's LINQ to objects.

Something like this?

```
Claim emailClaim = identity.Claims.SingleOrDefault(c => c.ClaimType == ClaimTypes.Email);
if (emailClaim != null)
{
    email = emailClaim.Value;   
}
```

davidfowl: Yes that looks better

woloski: Fixed on 77e0857

woloski:
Done with removing duplication on account linking into IdentityLinker service: eba63fa. Tests: 0d88209

davidfowl:
This guy needs an update (there's conflicting changes now).

brandwe:
I'll have this fixed by EOW

woloski:
This PR can be closed without merge, the rebase was more complex than I thought.
New one: #572

Issue #524, opened 2012-07-11
URL not rendering properly


lovatbrae:
URL's like this are not working: http://localhost/test

samandmoore:
this was fixed in #527

davidfowl:
"Fixed" :) It's still buggy as hell

PR #527, opened 2012-07-11
Simplified regex in TextTransform url parsing code

hasankhan:


hasankhan:
Also solves #524

JabbR/Infrastructure/TextTransform.cs
```diff
@@ -42,21 +42,21 @@ private string ConvertTextWithNewLines(string message)
...
+            message = urlPattern.Replace(message, m =>
             {
-
-                string httpPortion = String.Empty;
-                if (!m.Value.Contains("://"))
+                string url = m.Value;
+                if (!url.Contains("://"))
                 {
-                    httpPortion = "http://";
+                    url = "http://" + url;
```
davidfowl, line 55: This looks wrong. What about https and ftp etc? We need better unit tests.

hasankhan, line 55: This is saying if url does not have scheme, prefix it with http. Are you sure its a problem?

samandmoore, line 55: consensus, this looks good.

davidfowl, line 55: We should add a test for twitter urls.

hasankhan, line 55: @davidfowl  can you post a sample url?

samandmoore, line 55: https://twitter.com/jabbr/status/207283403274452992

samandmoore, line 55: we need to add a test that ensures that this fixes #524, can you add that @hasankhan ?

hasankhan, line 55: seems like its already done

PR #545, opened 2012-07-25
Improvements to date headers in chat

dpatti:
Some things I did:
- Fixed a bug where date header would sometimes not appear if you were in chat at the time of the day rollover
- Fixed an oddity where pulling in more history did not reset the date header at the top, which meant it would be left in between two messages of the same time
- Always show a date header at the very top
- User names always show next to messages after a date header
- The user name of the top message is removed during scrollback if the previous message was also by said user

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -1363,8 +1363,8 @@
             // If our top message is a date header, it might be incorrect, so we
             // check to see if we should remove it so that it can be inserted
             // again at a more apropriate time.
-            if ($target.data('timestamp')) {
-                var postedDate = new Date($target.data('timestamp')).toDate();
+            if ($target.is(".list-header.date-header")) {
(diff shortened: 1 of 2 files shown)


davidfowl:
send a screenshot of the changes

davidfowl:
Also Single quotes everywhere.

dpatti:
Excuse my lack of GIMP skills. This illustrates four of the points, the exception being the first which is hard to capture in a screenshot, but that is more of a bug and less of a feature.

Below, master is on the left and the branch changes are shown on the right. Blue lines are my emphasis. In case the second picture was not depicted well, the blue line is where the scrollback for the window stops, so when you scroll up you get a duplicate name on the left.

Date Header Changes

dpatti:
Fixed quotes and rebased onto current master. Let me know if you would like me to squash.

PR #546, opened 2012-07-25
Add click-to-message for users in chat and on list

dpatti:


JabbR/Chat.ui.js
```diff
@@ -847,6 +847,31 @@
                 }
             });
 
+            // handle click on names in chat / room list
+            var prepareMessage = function(ev) {
+                var message = $newMessage.val();
```
davidfowl: Don't replace it if theres's stuff there. If I click on a user by mistake I shouldn't lose my message

davidfowl:
Also can you fix the merge conflicts?

JabbR/Chat.ui.js
dpatti: It doesn't, if you already have some text in the box (assuming it isn't a slash command) it will prefix it with `/msg <username>`.

dpatti:
I rebased it onto your master and re-pushed.

davidfowl:
I can't auto merge from github. So there's conflicting changes.

JabbR/Chat.ui.js
davidfowl: You're right. 
You should trim the message value

dpatti:
Fixed double quotes and added trim. Still a bit puzzled that this is a merge conflict.

dpatti:
Oh, I see. Something is really messed up in my local repo with that recent ChatService.cs change. It somehow made its way into the commit I believe.

dpatti:
Okay, cleaned up the commit and rebased to the current master.

PR #547, opened 2012-07-25
Add basic tab cycling with keyboard

dpatti:
I've been finding this useful, but I'm not sure what your thoughts are. Basically, when you have no text in your input box, tab and shift+tab will cycle through your active chat rooms.


samandmoore:
I like it, but the last time I implemented this @davidfowl said he wasn't into it. I like your strategy better than what i did. Mine only cycled forward using shift+tab.

PR #552, opened 2012-07-27
Fix to user names after non-messages

dpatti:
In #545, a change was made that made timestamps show up correctly when non-chat
messages were entered before the chat was. As an unintended side-effect, the
user names were not being displayed after a non-chat message if that user spoke
before it, which made for some odd dialog.

To be more structured, all message appending was separated to a single function
regardless of message. There it is checked to see whether a new date header is
necessary or not. Now we have accurate date headers and user names appearing
when they should be.

Note that while no prepending code was touched, as it is not affected, it might
be something to refactor later.

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -1360,7 +1360,8 @@
         showGravatarProfile: function (profile) {
             var room = getCurrentRoomElements(),
                 nearEnd = ui.isNearTheEnd();
-            templates.gravatarprofile.tmpl(profile).appendTo(room.messages);
+
+            this.appendMessage(templates.gravatarprofile.tmpl(profile), room);
             if (nearEnd) {
                 ui.scrollToBottom();
             }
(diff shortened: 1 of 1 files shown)


JabbR/Chat.ui.js
```diff
@@ -1606,6 +1603,21 @@
 
             return $element;
         },
+        appendMessage: function (newMessage, room) {
+            // Determine if we need to show a new date header
+            if (!$(newMessage).is('.date-header')) {
+                var lastMessage = room.messages.find("li[data-timestamp]").last(),
```
davidfowl: single quotes

davidfowl:
Can you show what was broken and what this fixes? (visually)

dpatti:
Missing name

The second message here should be displaying a name for readability.

dpatti:
Here's another improvement: before date headers only showed before chat messages. As you can see here, a non-chat message happened when a date header should have been inserted. This was the primary reason for creating `appendMessage`.

No date header

davidfowl:
Since these changes seem to be regressing other features we'll need to find a way to add tests for these things.

PR #556, opened 2012-07-29
Improve the automatic scrolling for rich content

dpatti:
Three notable instances were corrected here:
- When someone enters an image link, once that image loads the window may scroll
  up past the bottom edge of the screen.
- When you join a room with images, as they load they will push the scroll up.
- When you paste content, it is later replaced with the full version from the
  server, which causes chat to be scrolled up.

The latter of the three was the most simple case, and it involved simply placing
the ui.replaceMessage() function into the scrollIfNecessary wrapper.

In the case of the images, there was already a system in place to force scroll
after 850ms in the case of a new message; however, this was not working since it
required you to already be at the end of the chat scroll. A new method was added
that can be called to attach load event handlers to elements within those
messages which will correct the scroll position only if the user himself has not
touched it in the meantime.

Deciding which elements to attach the load event handler to is done in the
utility function expandableContent(), which uses a hard-coded list of selectors
to search for within the messages provided. Most rich content in JabbR has
hard-coded dimensions, but the one exception that I found in my testing was
images for obvious reasons. If there are more, they can be easily added to the
array. At the very least, the functionality will not regress as it was never
working in the first place.

### DIFF
--- a/JabbR/Chat.js
+++ b/JabbR/Chat.js
@@ -65,9 +65,11 @@
                         ui.setRoomOwner(this, room);
                     });
 
+                    var messageIds = [];
                     $.each(roomInfo.RecentMessages, function () {
                         var viewModel = getMessageViewModel(this);
 
+                        messageIds.push(viewModel.id);
                         ui.addChatMessage(viewModel, room);
                     });
 
@@ -77,6 +79,7 @@
                     // that are added after initial population
                     ui.setInitialized(room);
                     ui.scrollToBottom(room);
+                    ui.watchMessageScroll(messageIds, room);
 
                     d.resolveWith(chat);
                 })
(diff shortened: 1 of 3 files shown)


JabbR/Chat.ui.js
```diff
@@ -1224,6 +1224,36 @@
                 room.scrollToBottom();
             }
         },
+        watchMessageScroll: function (messageIds, roomName) {
+            // Given a message id or array of message ids, if there is any
+            // embedded content in it, it may cause the window to scroll off of
+            // the bottom, so we can watch for that and correct it.
+            if (!(messageIds instanceof Array)) {
```
davidfowl: Instead of this, always pass an array.

JabbR/Chat.js
```diff
@@ -306,36 +309,25 @@
...
-        // Adding external content can sometimes take a while to load
-        // Since we don't know when it'll become full size in the DOM
-        // we're just going to wait a little bit and hope for the best :) (still a HACK tho)
-        window.setTimeout(function () {
-            var nearTheEndAfter = ui.isNearTheEnd(room);
-            if (nearTheEndBefore && nearTheEndAfter) {
-                ui.scrollToBottom();
-            }
-        }, 850);
+        ui.watchMessageScroll(id, room);
```
davidfowl:
```
ui.watchMessageScroll([id], room)
```

dpatti:
Fixed

JabbR/Chat.ui.js
```diff
@@ -1224,6 +1224,33 @@
                 room.scrollToBottom();
             }
         },
+        watchMessageScroll: function (messageIds, roomName) {
+            // Given an array of message ids, if there is any embedded content
+            // in it, it may cause the window to scroll off of the bottom, so we
+            // can watch for that and correct it.
+            messageIds = $.map(messageIds, function (id) { return '#m-' + id; });
+
+            var $messages = $(messageIds.join(',')),
```
davidfowl, line 1233: Didn't know you could do that $('#a1,#a2') returns a list of 2 jQuery elements?

dpatti, line 1233: Right, it would be the same as `$("#a1").add("#a2");`

JabbR/Chat.ui.js
```diff
@@ -1224,6 +1224,33 @@
...
+            if (nearTheEndBefore && $content.length > 0) {
+                // Note that the load event does not bubble, so .on() is not
+                // suitable here.
+                $content.load(function () {
+                    // If we used to be at the end and our scrollTop() did not
+                    // change, then we can safely call scrollToBottom() without
+                    // worrying about interrupting the user. We skip this if the
+                    // room is already at the end in the event of multiple
+                    // images loading at the same time.
+                    if (!room.messages.isNearTheEnd() && scrollTopBefore == room.messages.scrollTop()) {
```
davidfowl:
```
===
```

dpatti: Fixed

PR #576, opened 2012-08-11
HADDDOUUKEN - You know you want to

NTaylorMullen:
Secret command:
/hadouken


PureKrome:
LoL :thumbsup:

ajbeaven:
should be /ddff+p (Down, Down-Forward, Forward + Punch)

samandmoore:
Damian brought up a good point: this would be better if it worked like the nudge command. i.e. it needs to show up for all users in a specific room when the command is called.

NTaylorMullen:
Definitely a great plan =).  Not too hard to do either, especially seeing since a lot of the user friendly functionality is already there via nudge.

One issue though is getting images that are not copyrighted etc.  The sound is copywrite safe but the images I'll have to do some research into, I just got em off of Google images.

samandmoore:
@NTaylorMullen yeah, that's why it was never merged in the first place. I dunno enough about the copyright stuff to safely accept it ;)

PR #582, opened 2012-08-21
added Ban command

phillip-haydon:
Ok at long last I'm trying again.

This is what I was trying to comment last month, except now the ChatService no longer thinks every line changed. Using VS 2012 rather than 2010. That's the only difference.

Can now:

/ban _name_ 

Anyone, except me... :D

We can test by banning Daanl.


samandmoore:
what's with the ban user resx file?

phillip-haydon:
That is what was generated by the migration crap for EntityFailwork.

samandmoore:
@phillip-haydon i figured as much, but none of the other migrations generated that file. I'm guessing this has to do with VS2012. I guess we can just leave it.

phillip-haydon:
It's the first time I've done a migration with VS2012 so I don't know. 

I thought it was strange so I deleted the database and re-built and the database was created correctly so I assumed it was correct.

samandmoore:
we need to ensure that a user cannot log-in via the janrain route, not just via /nick

PR #587, opened 2012-09-02
Display the user's id when the /who command is invoked on own user

thomasvm:
I'm in the process of writing an app that can take the payload provided by GitHub's [post-receive hooks](https://help.github.com/articles/post-receive-hooks) and send the information into a Jabbr room.

To do so, I'm using the Jabbr.Client library which can authenticate using user/password combo, or using an id. Therefore it needs to be possible to expose your own userid.

Also doesn't render the gravatar info when the user's hash isn't set

### DIFF
--- a/JabbR/Chat.js
+++ b/JabbR/Chat.js
@@ -468,9 +468,11 @@
             ui.addMessage('Note: ' + userInfo.Note, 'list-item');
         }
 
-        $.getJSON('https://secure.gravatar.com/' + userInfo.Hash + '.json?callback=?', function (profile) {
-            ui.showGravatarProfile(profile.entry[0]);
-        });
+        if (userInfo.Hash) {
+            $.getJSON('https://secure.gravatar.com/' + userInfo.Hash + '.json?callback=?', function(profile) {
+                ui.showGravatarProfile(profile.entry[0]);
+            });
+        }
 
         chat.showUsersOwnedRoomList(userInfo.Name, userInfo.OwnedRooms);
     };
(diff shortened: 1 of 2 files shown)


davidfowl:
The user id is secret and exposing it via who is insecure. The user id is returned on login:

https://github.com/davidfowl/JabbR.Client/blob/master/JabbR.Client/Models/LogOnInfo.cs

thomasvm:
Fair enough, any alternatives you can think of? Generating a revokable auth token or something like that?

davidfowl:
You get access to the user id when you login, what more do you need? I guess I'm not understanding the scenario...

thomasvm:
Well, I want to perform the authentication from a server application, where we store the authentication (user id or username/password) in a configuration file. I'm using your JabbR.Cliënt library which provides two ways of doing the authentication
- username/password, which works fine if JabbR is configured without janrain
- user id, which is what we'll need when janrain is enabled because then username/password doesn't make any sense.

So what I want is that a user can discover his own user id, so he can use that to setup the server app.

I hope this clarifies things a bit.

davidfowl:
How are you going to authenticate with Janrain from a non webapp? I'm still not understanding things clearly. If you want to get your user id from jabbr itself look in the the cookie called jabbr.state. Or look in the database, it really isn't that hard. I'm fine adding a command that shows the user id for the logged in user but in a very secret way since there's no way to change it right now. So the ui should be such hidden we should probably be able to delete it from my local chat window.

thomasvm:
Thanks for your patience, impressive :-)

You're probably right, it's  better to leave things as they are and not expose this functionality through the UI. I know it's easy to look into the database or into the cookie, and for my use case it's more than enough.

Anyway, if you would like this see how this information is used, check out [this repo](https://github.com/thomasvm/github-jabbr). To be more precise, in this piece of code
https://github.com/thomasvm/github-jabbr/blob/master/Jabbr.GitHub/App/JabbrSend.cs#L72

Issue #580, opened 2012-08-21
"No users" missing under Online and Owners headings


nschonni:
When a user changes from Online to Away, or an Owner leaves a room, the "No Users" message should be shown if there are no other users under that section. It appears the function that adds the No User message https://github.com/davidfowl/JabbR/blob/master/JabbR/Chat.ui.js#L301 is not invoked.

WilliamBZA:
Should be sorted with PR: https://github.com/davidfowl/JabbR/pull/591

Issue #558, opened 2012-07-31
on-load away users should be faded


samandmoore:
when you load a room or log-on, the user list should show away users as grayed out.

this behavior should mimic what happens when a user goes from active to away while you are online.

samandmoore:
need to repurpose this https://github.com/davidfowl/JabbR/blob/master/JabbR/Chat.ui.js#L1339

WilliamBZA:
Should be sorted with PR: https://github.com/davidfowl/JabbR/pull/591

PR #591, opened 2012-09-26
Fix for Issue #580 and issue #558

WilliamBZA:
"No users" missing under Online and Owners headings:

Made sure that the chat.ui setListState method was called when an owner leaves the room or when a user joins a room with no owners currently inside.


WilliamBZA:
Also added fix for 558. Inactive users are now faded in 1ms for new users when the new users join the room.

Issue #518, opened 2012-06-27
Urban Dictionary Command Provider


phillip-haydon:
The ability to type:

/urban [word]

And have the first result of urban dictionary embedded into the chat. 

Possibly include #

/urban [word] [entry #]

To embed a certain entry in urban dictionary.

WilliamBZA:
Don't think it makes sense to have this as a command, should rather be a ContentProvider.

I've knocked up a sample here: https://github.com/WilliamBZA/JabbR/blob/master/JabbR/ContentProviders/UrbanDictionaryContentProvider.cs

phillip-haydon:
When I wrote the issue I think I meant Content Provider but I was probably suffering from lack of caffeine :)

PR #593, opened 2012-09-27
Add UrbanDictionary content provider

WilliamBZA:
To address this issue: https://github.com/davidfowl/JabbR/issues/518

Standard approach, get parameters, make request to API and render result.

### DIFF
--- a/JabbR/ContentProviders/UrbanDictionaryContentProvider.cs
+++ b/JabbR/ContentProviders/UrbanDictionaryContentProvider.cs
@@ -1,0 +1,77 @@
+﻿using System;
+using System.Linq;
+using System.Threading.Tasks;
+using JabbR.ContentProviders.Core;
+using JabbR.Infrastructure;
+using System.Text.RegularExpressions;
+using SignalR.Hosting.Common;
+
+namespace JabbR.ContentProviders
+{
+    public class UrbanDictionaryContentProvider : CollapsibleContentProvider
+    {
+        private static readonly string _contentFormat = @"
+<article>
+    <div>
+        <a href=""{0}"" target=""_blank"">{1}</a>
+    </div>
+    <b>Definition:</b>
+    <div style=""margin-left:1.5em;margin-bottom:1em;"">{2}</div>
+    <b>Usage:</b>
+    <div style=""font-style: italic; margin-left:1.5em;"">{3}</div>
+</div>";
+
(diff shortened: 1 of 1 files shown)


JabbR/ContentProviders/UrbanDictionaryContentProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+                return null;
+            }
+
+            var apiQuery = string.Format(_urbanDictionaryAPIURL,
+                string.IsNullOrWhiteSpace(term) ? string.Empty : string.Format("term={0}", term),
+                string.IsNullOrWhiteSpace(defid) ? string.Empty : string.Format("&defid={0}", defid));
+
+            return FetchArticle(new Uri(apiQuery)).Then(result =>
+                {
+                    var list = result.list;
```
samandmoore, line 42: Do you need to do a check here to verify that the result came back with data, or is that handled implicitly somewhere in the core code? I can't remember.

samandmoore:
Do you think you could squash your commits before we merge this so we don't get that hideous merge commit message?

JabbR/ContentProviders/UrbanDictionaryContentProvider.cs
WilliamBZA, line 42: Yeah, it breaks with not found items. Will fix and resubmit sometime.

JabbR/ContentProviders/UrbanDictionaryContentProvider.cs
```diff
@@ -0,0 +1,74 @@
...
+            string term = parameters["term"];
+
+            if (string.IsNullOrWhiteSpace(term) && string.IsNullOrWhiteSpace(defid))
+            {
+                return null;
+            }
+
+            var apiQuery = string.Format(_urbanDictionaryAPIURL,
+                string.IsNullOrWhiteSpace(term) ? string.Empty : string.Format("term={0}", term),
+                string.IsNullOrWhiteSpace(defid) ? string.Empty : string.Format("&defid={0}", defid));
```
davidfowl, line 38: Big S for String. methods

WilliamBZA, line 38: Just the methods or the variables too? (private static readonly string _contentFormat = blah vs private static readonly String _contentFormat = blah)

davidfowl, line 38: Just the methods

WilliamBZA:
Will resubmit when I get time again

PR #601, opened 2012-11-08
Use sparse ViewModel for lobby JSON

nschonni:
Currently tabbing into the lobby causes a >100KB JSON payload.
This can be reduced by removing properties not used by the Lobby:
- Topic
- Users
- Welcome
- Owners
- Recent Messages

### DIFF
--- a/JabbR/Hubs/Chat.cs
+++ b/JabbR/Hubs/Chat.cs
@@ -265,12 +265,12 @@ namespace JabbR
             };
         }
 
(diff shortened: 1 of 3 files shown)


nschonni:
From @davidfowl check to see if JSON.Net attributes could be used instead of a separate viewmodel

nschonni:
Looked at http://stackoverflow.com/questions/6507889/how-to-ignore-a-property-in-class-if-it-is-null-using-json-net but I think the separate model is a safer approach

PR #602, opened 2012-11-08
Update SquishIt to 0.9.1

nschonni:
0.9 branch removes the dependencies on the LESS and other preprocessors that JabbR doesn't use


nschonni:
Screwed up the rebase, will resubmit

PR #604, opened 2012-11-08
Update SquishIt to 0.9.1

nschonni:
Fixup for bad rebase in #602
Also dependant AjaxMin and YUI.Compressor.net packages


davidfowl:
Did you test it? What;s new. I see some missing config sections

nschonni:
Yeah, SquishIt just moved the dependencies for LESS and Sass compilation to separate packages. Now dotless and the Jurassic aren't required because JabbR doesn't uses LESS or Sasss.

PR #606, opened 2012-11-10
Cleanup unused js

nschonni:
Remove unrefereneced scripts and packages


JabbR/JabbR.csproj
```diff
@@ -820,7 +819,6 @@
     <Content Include="Scripts\jquery.timeago.0.10.js" />
     <Content Include="Scripts\jQuery.tmpl.js" />
     <Content Include="Scripts\jQuery.tmpl.min.js" />
-    <Content Include="Scripts\jquery.writeCapture.js" />
```
davidfowl, line 823: You sure this isn't used?

samandmoore, line 823: I was pretty sure this was used by one of the content providers.

nschonni, line 823: I couldn't find anything referencing it. I believe they all use jquery.captureDocumentWrite.js now

samandmoore, line 823: Ah. Right. that makes sense. Well, as long as you've verified things.

samandmoore, line 823: @nschonni pretty sure you're right on that one. lgtm.

PR #608, opened 2012-11-10
Toast fallback toastr

nschonni:
In browsers that don't support webkitNotifications, it adds a peek pop-up inside the JabbR window

Issue #605, opened 2012-11-09
Add Toast fallback for non-webkit browsers


nschonni:
I was thinking this might work https://github.com/CodeSeven/toastr but it is in browser only

davidfowl:
Nah

PR #609, opened 2012-11-10
Toast fallback toastr

nschonni:
For #605
In browsers that don't support webkitNotifications, it adds a peek pop-up inside the JabbR window using [toastr](https://github.com/CodeSeven/toastr)

Rebased version of #608

### DIFF
--- a/JabbR/Chat.css
+++ b/JabbR/Chat.css
@@ -104,7 +104,6 @@
   height: 20px;
   left: 55px;
   position: absolute;
-  display: none;
 }
 
   #preferences .toast.off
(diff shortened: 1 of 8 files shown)


JabbR/Chat.css
```diff
@@ -104,7 +104,6 @@
   height: 20px;
   left: 55px;
   position: absolute;
-  display: none;
```
davidfowl, line 107: why?

nschonni, line 107: Because now we don't have to hide the button, all browser now (kinda) support the notifications

samandmoore, line 107: so then are we saying that in-browser-window toast feature === desktop-notification-toast?

nschonni, line 107: no, just a fallback. you can try it out http://jabbr-nick.apphb.com/#/rooms/toast

davidfowl:
Can you rebase and update against latest?

nschonni:
I already rebased it, what's confilicting?

davidfowl:
Overall this isn't useful because the notifications only live within jabbr itself.

PR #616, opened 2013-01-21
Set Null initializer for EF and use LocalDb

samandmoore:
set null initializer for entity framework to prevent weird startup issues when db doesn't exist and use local db instead of sql express

@davidfowl any objections?

I noticed that without the null initializer, EF was creating the db for me then failing to run the migrations because the schema was created based on the newer model (very odd).

and I think we should just use localdb :smile:

### DIFF
--- a/JabbR/Models/JabbrContext.cs
+++ b/JabbR/Models/JabbrContext.cs
@@ -10,6 +10,11 @@ namespace JabbR.Models
         {
         }
 
+        static JabbrContext()
+        {
+            Database.SetInitializer<JabbrContext>(null);
+        }
+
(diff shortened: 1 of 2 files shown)


davidfowl:
Yes, I don't want to use local db by default as it doesn't work in IIS. Also I'm not sure why we need to set a null initializer.

samandmoore:
In order to get the database to create without throwing SQL errors I had to set the initializer to null. I think it's because we have an initial migration that creates all the tables, but the default EF initializer already does that. So then everything explodes trying to migrate. I will remove the localdb bits.

samandmoore:
closing, will do manually.

PR #618, opened 2013-01-24
Implement new login screen

hnrkndrssn:


JabbR/Nancy/AuthenticationModule.cs
```diff
@@ -24,6 +24,8 @@ public AuthenticationModule(IAuthenticationService authService, IMembershipServi
                 return 403;
             };
 
+            Get["/login"] = _ => View["/login"];
```
davidfowl: Can't you just do View["login"]

JabbR/Chat.js
```diff
@@ -942,7 +942,7 @@
                 .fail(function (e) {
                     // Not authorized so show the login screen
                     if (e.status === 403) {
-                        ui.showLogin();
+                        window.location = '/auth/login';
```
davidfowl: Don't make it absolute. Do:

``` javascript
window.location = 'auth/login';
```

PR #623, opened 2013-01-25
Add support to enable the flags of British home nations

DavidChristiansen:
Utilised non ISO standard codes to enable the flags for British home nations

G1 - England,
G2 - Wales,
G3 - Scotland

### DIFF
--- a/JabbR/Chat.css
+++ b/JabbR/Chat.css
@@ -1187,7 +1187,7 @@ li.clearfix
     background-position: -208px -33px;
   }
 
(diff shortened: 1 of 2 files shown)


JabbR/Chat.css
```diff
@@ -1837,7 +1837,7 @@ li.clearfix
     background-position: -240px -121px;
   }
 
-  .flag.flag-scotland
+  .flag.flag-zz
```
davidfowl: I don't get this change

PR #637, opened 2013-01-26
Layout changes

hnrkndrssn:
Aligning registration page when there are no social providers
Make profile page pretty and stuff


JabbR/Views/Account/login.cshtml
```diff
@@ -1,19 +1,29 @@
...
 }
 @switch ((AuthenticationMode)Model.AuthenticationMode)
 {
     case AuthenticationMode.Default:
     <div class="row-fluid">
-        <div class="span5">
+      <div class="span5 @if (!hasSocialProviders)
+                        {
+                          <text>offset3</text>
+                        }">
```
samandmoore: can we get this all onto 1 line like below? :smile:

PR #638, opened 2013-01-26
ensure that provider and identity are unique in the database

### LINKED ISSUE #634: Ensure a no dupes for an (identity, providerid) pair in the ChatUserIdentity table

samandmoore:
- prevent sign-in / user-creation when providerName & identity pair are already present
- fixed #634 

@davidfowl peep it.

i guess there is probably something in here that we can test, but i don't have time to write the tests _right now_


davidfowl:
We just want to prevent user creation since this should never happen in normal login/creation code path.

samandmoore:
yeah, i think that's what it does.

davidfowl:
Merging in

davidfowl:
Here's what I get from creating the index 

```
Column 'Identity' in table 'dbo.ChatUserIdentities' is of a type that is invalid for use as a key column in an index.
```

Issue #624, opened 2013-01-25
Add form field validation for username/password auth.


davidfowl:
There's no validation messages shown today when login or user creation fails.

davidfowl:
Fixed just needs to not look like shit now :smile:

PR #640, opened 2013-01-27
added the ability to display error messages using html helpers

samandmoore:
what you've got here is a few html helpers and a base module class that add the ability to add error messages to your forms. booyah.

@davidfowl


samandmoore:
basically fixes #624

davidfowl:
w00T! Just needs to be merged in but there's conflicts.

PR #660, opened 2013-01-28
Fixing userlist-header

Buildstarted:
This is to remove the extraneous space after the userlist header so that
the lists aren't so separated from their headers.


davidfowl:
Image please.

Buildstarted:
original
fixed

PR #691, opened 2013-02-10
Fixed #689

### LINKED ISSUE #689: updating your flag doesn't update the image
If you already have your flag set and then change it the tooltip updates in real-time but the image stays as the old value. You need to refresh the site for the new image to show.

schulz3000:
removeClass() won't work so used removeAttr('class') instead


davidfowl:
Why doesn't it work?

schulz3000:
I don't know, looks like a problem in jQuery. I write a jsFiddle http://jsfiddle.net/h3jCk/1/ for this. If you give the class as parameter to removeClass it's working but parameterless removeClass won't work. Also google gives me no answer for this.

davidfowl:
Your other changes are in this PR

PR #701, opened 2013-02-27
Added infrastructure for reload of messages after disconnect

hnrkndrssn:
The result after calling the new event for displaying a reload messages notification
ReloadMessage

### DIFF
--- a/JabbR/Chat.js
+++ b/JabbR/Chat.js
@@ -909,6 +909,10 @@
             });
     });
 
+    $(ui).bind(ui.events.reloadMessages, function (ev, roomInfo) {
+        ui.showReloadMessage(roomInfo.name);
+    });
+
     $(ui).bind(ui.events.preferencesChanged, function (ev) {
         updateCookie();
     });
--- a/JabbR/Chat.js
+++ b/JabbR/Chat.js
@@ -910,7 +910,7 @@
     });
 
     $(ui).bind(ui.events.reloadMessages, function (ev, roomInfo) {
-        ui.showReloadMessage(roomInfo.name);
+        ui.reloadMessages(roomInfo);
     });
 
     $(ui).bind(ui.events.preferencesChanged, function (ev) {
--- a/JabbR/Chat.js
+++ b/JabbR/Chat.js
@@ -909,7 +909,7 @@
             });
     });
 
-    $(ui).bind(ui.events.reloadMessages, function (ev, roomInfo) {
+    $(ui).bind(ui.events.reloadMessages, function () {
 
     });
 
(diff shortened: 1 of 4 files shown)


JabbR/Chat.js
```diff
@@ -909,6 +909,10 @@
             });
     });
 
+    $(ui).bind(ui.events.reloadMessages, function (ev, roomInfo) {
+        ui.showReloadMessage(roomInfo.name);
```
davidfowl: What is this?

davidfowl: Who triggers this?

hnrkndrssn: That would be called when it is detected that a room is missing messages due to a disconnection (for example after resuming computer from sleep).

Or have I assumed incorrectly?

Sent from my Windows Phone

davidfowl:
Shouldn't that get triggered when clicking on the reload button?

hnrkndrssn:
Oh, sorry I misunderstood then. That event is just for displaying that notification.

I didn't do the event for actually reloading the messages. I can add that tonight.

Sent from my Windows Phone

hnrkndrssn:
And I will work on the naming of the events as well...:|

Sent from my Windows Phone

JabbR/Chat.ui.js
```diff
@@ -1910,6 +1929,9 @@
             },
             updateTimeout);
         },
+        reloadMessages: function(roomName) {
```
davidfowl: Remove this.

JabbR/Chat.js
```diff
@@ -909,6 +909,10 @@
             });
     });
 
+    $(ui).bind(ui.events.reloadMessages, function (ev, roomInfo) {
+        ui.reloadMessages(roomInfo);
```
davidfowl: Remove this.

JabbR/Chat.ui.js
```diff
@@ -784,6 +785,16 @@
         }, cycleTimeInMilliseconds);
     }
 
+    function showReloadMessageNotification(roomName) {
+        var room = getRoomElements(roomName);
+        if (!room.isLobby()) {
```
davidfowl: What happens if detection happens and I'm on the lobby? It should always show up. I'm not sure I want to read messages per room.

hnrkndrssn: OK, so you want the reload to reload all messages in all open rooms? In that case I'll remove the room name params

JabbR/Chat.ui.js
```diff
@@ -1910,6 +1917,10 @@
             },
             updateTimeout);
         },
+        showReloadMessageNotification: function () {
+            $reloadMessageNotification.appendTo($chatArea);
```
davidfowl, line 1921: If I call this mutliple times does it append multiple elements? Should there be a hide?

hnrkndrssn, line 1921: I have tested calling this multiple times from different areas in the code and it only ever adds the one element.

Issue #677, opened 2013-01-30
Profile page messed up for users with really long usernames


hnrkndrssn:
![profile-page-really-long-username](https://f.cloud.github.com/assets/1035315/110529/6995d13a-6ac4-11e2-8245-959b5445d4cd.png)

PR #703, opened 2013-03-08
Username positioning on preferences screen

cmaish:
Minor change to make account gravatar/username inline elements, handle loading positioning and overflow with imagesize and elipses, fixing #677


cmaish:
Before:
677-Before

After:
677-After

PR #721, opened 2013-03-24
Fix up the user list sorting, adding them to list with correct status

hnrkndrssn:
Also removed unused idle list as all users are now added to the active users list


davidfowl:
Looks good.

PR #741, opened 2013-03-26
2 changes to CommandManager

philchuang:
1: case-insensitive commands
2: invalid /commands are now exceptions reported to the user and not displayed to the chat room.

### DIFF
--- a/JabbR.Tests/CommandManagerFacts.cs
+++ b/JabbR.Tests/CommandManagerFacts.cs
@@ -2654,6 +2654,171 @@ namespace JabbR.Test
 
                 Assert.Throws<InvalidOperationException>(() => commandManager.TryHandleCommand("/invite dfowler"));
             }
+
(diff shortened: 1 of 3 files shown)


davidfowl:
Not sure how I feel about this change. Maybe it should only fail for certain commands.

philchuang:
IMHO, if I'm entering a / command then I also don't want to broadcast it to the room. Just my preference, though. 

What's the use case for sending unrecognized commands to the room? when someone wants to type a literal /something? In that case an exception could be made for sending literal leading forward slashes.

I'm trying to remember how it was back in my IRC days, I think it was up to each server's implementation to decide, but IIRC they usually reported back to the user that it was an invalid command and did not post such to the room.

philchuang:
ok, i'm deleting my fork and re-forking w/ branches this time for more focused PRs

PR #742, opened 2013-03-26
Sort user list, split the users into active/idle then sort each group separately

hnrkndrssn:
Sending another attempt at sorting the userlist by status and name for review.
I'm not sure of the efficiency of this script when lots of activity is going on but let me know your concerns and I will attempt to justify/rectify them.

I created a JsFiddle to attempt to prove that it works.

http://jsfiddle.net/n2gNC/10/

Uncomment the two calls to sortList and then click Run and wait for the list to be sorted (takes a little while)

Then you can click a user in the list, if the user is active they will be marked as idle and vice versa.

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -410,8 +410,6 @@
                     idleUsers.push(item);
                 }
             });
-            console.log(activeUsers.length);
-            console.log(idleUsers.length);
 
             activeUsers.sort(function (a, b) {
                 var compA = $(a).data('name').toString().toUpperCase();
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -392,6 +392,14 @@
             }
         };
 
+        this.sortUsersByName = function(userListToSort) {
+            return userListToSort.sort(function(a, b) {
+                var compA = $(a).data('name').toString().toUpperCase();
+                var compB = $(b).data('name').toString().toUpperCase();
+                return (compA < compB) ? -1 : (compA > compB) ? 1 : 0;
+            });
+        };
+
         this.sortLists = function () {
             this.sortList(this.owners);
             this.sortList(this.activeUsers);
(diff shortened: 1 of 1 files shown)


samandmoore:
What about when new users join and current users leave?

hnrkndrssn:
Have added a call to the sortlist function when adding a new user, when a user leaves, the list will still stay sorted as they are merely removed from the list.

I have also updated the fiddle http://jsfiddle.net/n2gNC/12/

JabbR/Chat.ui.js
```diff
@@ -392,21 +393,46 @@
...
+                idleUsers = [];
+
+            $.each(listItems, function (index, item) {
+                if ($(item).data('active')) {
+                    activeUsers.push(item);
+                } else {
+                    idleUsers.push(item);
+                }
+            });
+            console.log(activeUsers.length);
```
davidfowl: Get rid of these

samandmoore:
this looks good to me

JabbR/Chat.ui.js
```diff
@@ -392,21 +393,46 @@
...
+                var compB = $(b).data('name').toString().toUpperCase();
+                return (compA < compB) ? -1 : (compA > compB) ? 1 : 0;
+            });
+            
+            $.each(activeUsers, function (index, item) {
+                sortedUsers.push(item);
+            });
+            $.each(idleUsers, function (index, item) {
+                sortedUsers.push(item);
+            });
```
samandmoore: instead of this can't you do

```
sortedUsers = activeUsers.concat(idleUsers);
```

JabbR/Chat.ui.js
```diff
@@ -392,21 +393,46 @@
...
-                var compA = $(a).data('active');
-                var compB = $(b).data('active');
-                return (compA > compB) ? -1 : (compA < compB) ? 1 : 0;
             });
-            $.each(listItems, function (index, item) { listToSort.append(item); });
+            idleUsers.sort(function (a, b) {
+                var compA = $(a).data('name').toString().toUpperCase();
+                var compB = $(b).data('name').toString().toUpperCase();
+                return (compA < compB) ? -1 : (compA > compB) ? 1 : 0;
+            });
```
samandmoore: so the code within these two separate sort calls is identical right? sounds like we should define this as a `sortUsersByName` function and re-use it

hnrkndrssn:
I have updated the code as per @samandmoore's comments :smile:

davidfowl:
Then youre good to merge

PR #743, opened 2013-03-26
Adding link to surround image

Buildstarted:
Proposing that images should be surrounded by anchor tags to allow easy linking to full versions


davidfowl:
Totally misread the bug (forcing you to do it yourself).

Issue #745, opened 2013-03-27
File upload is still available on lobby and closed rooms (and works on closed rooms)


cmaish:
The file upload button is not disabled in closed rooms.  File upload will work, and the file is added to the message history for the room (which can be verified by downloading a list of messages for the room).

![jabbr-room-file-upload](https://f.cloud.github.com/assets/1271535/306859/ccf1e250-969d-11e2-99ed-557d0ff55eed.png)

PR #746, opened 2013-03-27
Disable file upload on lobby and in closed rooms

cmaish:
The file upload button (and backing file upload control) is not disabled in the lobby and closed rooms.

File uploads succeed in closed rooms (though changes are not displayed) and the upload seems to be added to the room history.

Should fix issue #745

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -2066,11 +2066,13 @@
                 $hiddenFile.attr('disabled', 'disabled');
                 $submitButton.attr('disabled', 'disabled');
                 $newMessage.attr('disabled', 'disabled');
+                $fileUploadButton.attr('disabled', 'disabled');
             }
             else {
                 $hiddenFile.removeAttr('disabled');
                 $submitButton.removeAttr('disabled');
                 $newMessage.removeAttr('disabled');
+                $fileUploadButton.removeAttr('disabled');
             }
         },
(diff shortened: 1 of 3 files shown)


JabbR/Chat.ui.js
```diff
@@ -2131,12 +2133,11 @@
         },
         shouldCollapseContent: shouldCollapseContent,
         collapseRichContent: collapseRichContent,
-        toggleMessageSection: function (disabledIt) {
-            if (disabledIt) {
-                // disable button and textarea
+        toggleMessageSection: function (disableEdit, disableFileUpload) {
```
davidfowl: Why do you need both flags? Don't they always go together?

cmaish: I originally threw everything under the one argument, but the lobby allows messages but doesn't do anything on file uploads - happy to recombine though

davidfowl: The lobby allows commands not messages. I'd rather have it under a single flag.

cmaish:
Altered to keep a single flag on toggleMessageSection.  A side effect of this is that the lobby doesn't disable the file upload button at current.  Slightly altered ordering in UploadCallbackHandler to give useful messages when the message can't be added to the room for some reason.

jabbr-file-upload-lobby

Does anything need to be done with the lobby?  I could see it being useful for the lobby's button to just return the url of the uploaded file (for things like PMs)...

PR #747, opened 2013-03-27
When the last user exits a room, show 'no users' to remaining room owners

cmaish:
When the last non-room-owner user exits a room, the users menu does not currently update to show "No users".

jabbr-room-last-user-leaves

This functionality does work when the last owner leaves.


samandmoore:
looks good.

PR #748, opened 2013-03-27
case-insensitive command matching

philchuang:
/Create = /CrEaTe = /create, etc

### DIFF
--- a/JabbR/Commands/CommandManager.cs
+++ b/JabbR/Commands/CommandManager.cs
@@ -148,7 +148,7 @@ namespace JabbR.Commands
 
             IList<string> candidates = null;
 
-            var exactMatches = _commandCache.Keys.Where(comm => comm.Equals(commandName))
+            var exactMatches = _commandCache.Keys.Where(comm => comm.Equals(commandName, StringComparison.CurrentCultureIgnoreCase))
                                                  .ToList();
 
             if (exactMatches.Count == 1)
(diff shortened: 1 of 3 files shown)


JabbR/JabbR.csproj
```diff
@@ -22,9 +22,9 @@
     <OldToolsVersion>4.0</OldToolsVersion>
     <UpgradeBackupLocation />
     <IISExpressSSLPort />
-    <IISExpressAnonymousAuthentication />
-    <IISExpressWindowsAuthentication />
-    <IISExpressUseClassicPipelineMode />
+    <IISExpressAnonymousAuthentication>enabled</IISExpressAnonymousAuthentication>
```
davidfowl: This shouldn't be in this change.

philchuang: yup, noticed that too late, will revert

JabbR/Commands/CommandManager.cs
```diff
@@ -157,7 +157,7 @@ select new
             }
             else
             {
-                candidates = _commandCache.Keys.Where(comm => comm.StartsWith(commandName))
+                candidates = _commandCache.Keys.Where(comm => comm.StartsWith(commandName, StringComparison.CurrentCultureIgnoreCase))
```
davidfowl: OrdinalIgnoreCase

samandmoore:
what @davidfowl said, then looks good to me

PR #749, opened 2013-03-27
CommandManager now throws on unmatched command

philchuang:
b/c why should the chat room see invalid or mistyped commands?


samandmoore:
seems like a good change.

this will prevent me from embarrassing myself when i mistype /mssg

PR #750, opened 2013-03-27
InviteCommand now works with multiple users

philchuang:
like /invite user1 user2 user3 room

### DIFF
--- a/JabbR/Commands/InviteCommand.cs
+++ b/JabbR/Commands/InviteCommand.cs
@@ -1,25 +1,18 @@
 ﻿using System;
+using System.Collections.Generic;
+using System.Linq;
 using JabbR.Models;
 
 namespace JabbR.Commands
 {
-    [Command("invite", "Invite a user to join a room.", "user room", "room")]
+    [Command("invite", "Invite a user to join a room.", "user [more users] room", "room")]
     public class InviteCommand : UserCommand
     {
         public override void Execute(CommandContext context, CallerContext callerContext, ChatUser callingUser, string[] args)
         {
             if (args.Length == 0)
             {
-                throw new InvalidOperationException("Who do you want to invite?");
-            }
-
-            string targetUserName = args[0];
-
-            ChatUser targetUser = context.Repository.VerifyUser(targetUserName);
-
-            if (targetUser == callingUser)
-            {
-                throw new InvalidOperationException("You can't invite yourself!");
+                throw new InvalidOperationException("Who do you want to invite, and into what room?");
             }
 
             if (args.Length == 1)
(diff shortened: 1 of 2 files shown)


JabbR/Commands/InviteCommand.cs
```diff
@@ -1,36 +1,62 @@
...
+            {
+                throw new InvalidOperationException("Invite them to which room?");
+            }
 
-            ChatUser targetUser = context.Repository.VerifyUser(targetUserName);
+            String roomName = null;
+            var targetUsers = new List<ChatUser>();
 
-            if (targetUser == callingUser)
+            for (int i = 0; i < args.Length; i++)
```
davidfowl, line 26: Loop to args.Length - 1 since the last arg is the room.

JabbR/Commands/InviteCommand.cs
```diff
@@ -1,36 +1,62 @@
...
+                if (String.Equals (targetUserName, callingUser.Name, StringComparison.CurrentCultureIgnoreCase))
+                {
+                    throw new InvalidOperationException("You can't invite yourself!");
+                }
+
+                try
+                {
+                    targetUsers.Add (context.Repository.VerifyUser(targetUserName));
+                }
+                catch (InvalidOperationException)
```
davidfowl, line 40: Get rid of this catch.

JabbR/Commands/InviteCommand.cs
```diff
@@ -1,36 +1,62 @@
...
+            if (roomName == null)
             {
                 throw new InvalidOperationException("Invite them to which room?");
             }
 
-            string roomName = args[1];
             ChatRoom targetRoom = context.Repository.VerifyRoom(roomName);
 
-            context.NotificationService.Invite(callingUser, targetUser, targetRoom);
+            foreach (var targetUser in targetUsers)
```
davidfowl, line 58: Always add braces.

JabbR/Commands/InviteCommand.cs
philchuang, line 40: I did it this way so that if a user forgot to specify a room, like
"/invite user1 user2 user3"
the command would at least verify all the given usernames.

If it was automatically assumed the last parameter was a room, then user1 & user2 would get an invite into room "user3".

davidfowl, line 40: Still don't like it

philchuang, line 40: scenario: albert types /invite bonnie charles david

method 1: always assume last argument is a room

if room david exists,
then bonnie and charles are invited to room david

if room david does not exist,
then albert is told "unable to find room david"

method 2: check to see if every argument is a user

if user david exists,
then albert is told "invite them to which room?"

if user david does not exist, and room david exists, 
then bonnie and charles are invited to room david

if user david does not exist, and room david does not exist,
then albert is told "unable to find room david"

Neither method is perfect, since with multiple users that last parameter is awfully ambiguous. But I believe method 2 is perhaps more insightful in its error messages.

But I can think of one way to make this multi-user-invite completely un-ambiguous, and that would be to require the # in front of the room name.

So when albert types "/invite bonnie charles david" the system will correctly respond "invite them to which room?", regardless of the existence of user david or room david.

The correct usage of the invite command would then be, "/invite bonnie charles david #abcd".

This is consistent with how roomnames are displayed throught Jabbr - however it represents a big change in the invite command.

Thoughts?

PR #763, opened 2013-04-02
message history trimming on timer

samandmoore:
uses a timer to trim the message history for the currently open rooms.
- skips lobby
- only trims rooms that have been typed in since last history fetch
- trims messages exceeding 500

@davidfowl @alfhenrik

### DIFF
--- a/JabbR/Chat.js
+++ b/JabbR/Chat.js
@@ -767,6 +767,7 @@
             typing = true;
 
             try {
+                ui.setRoomTrimmable(chat.state.activeRoom, typing);
                 chat.server.typing(chat.state.activeRoom);
             }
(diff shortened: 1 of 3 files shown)


JabbR/Chat.js
```diff
@@ -973,6 +975,10 @@
         logout();
     });
 
+    $(ui).bind(ui.events.trimMessageHistory, function(ev, roomName) {
```
davidfowl: What's the point having an event if all it does it call back into ui.trimMessageHistory

samandmoore: i was originally going to do more in the event handler. we can probably remove it now. that's why i need you to review ;)

PR #764, opened 2013-04-02
GithubIssuesCommentContentProvider.cs

Buildstarted:
- Added template to index.cshtml
- Added javascript callback for api call
- Added provider to ContentProviders

### DIFF
--- a/JabbR/Chat.githubissues.js
+++ b/JabbR/Chat.githubissues.js
@@ -30,23 +30,8 @@
         }
     };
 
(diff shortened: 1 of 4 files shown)


JabbR/Chat.githubissues.js
```diff
@@ -30,6 +30,46 @@
         }
     };
 
+    window.addGitHubIssueComment = function () {
+        //github-issues-comment-template
+        var comment = {
```
davidfowl: I assume this is unneeded?

PR #765, opened 2013-04-02
First message tag and CSS styling.

DavidChristiansen:
# Full Width

# 1st width reduction

# 2nd width reduction

## Note:

This PR is the first of a few planned PRs to introduce better responsive layouts for smaller screens. 
The navigation, user list and message entry controls all need addressed

### DIFF
--- a/JabbR/Chat.css
+++ b/JabbR/Chat.css
@@ -5,7 +5,6 @@
 }
 
 input, button, select, textarea {
-    font-family: Lucida Grande, Trebuchet MS;
     font-size: 13px;
 }
 
@@ -140,7 +139,6 @@ form#send-message {
 
     #heading h1 {
         margin-left: 5px;
-        font-family: 'Lucida Grande', 'Trebuchet MS';
         line-height: normal;
         font-size: 20px;
     }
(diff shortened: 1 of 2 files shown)


JabbR/Chat.css
```diff
@@ -140,7 +139,6 @@ form#send-message {
 
     #heading h1 {
         margin-left: 5px;
-        font-family: 'Lucida Grande', 'Trebuchet MS';
```
DavidChristiansen, line 143: Not necessary to repeat of font family declaration.

JabbR/Chat.css
```diff
@@ -526,7 +524,6 @@ li.room {
 }
 
 #new-message {
-    font-family: Trebuchet MS;
```
DavidChristiansen, line 529: Inconsistent font-family declaration

JabbR/Chat.css
```diff
@@ -232,7 +230,7 @@ form#send-message {
 }
 
 .user .details .name {
-    max-width: 70px;
+    max-width: 100px;
```
hnrkndrssn: I have a feeling this will cause some wrapping of the users if a user is an Admin, has a flag and is Idle (ie Idle Timestamp is displayed), please add a screenshot of this scenario.

PR #775, opened 2013-04-03
increase a few message counts

samandmoore:
@davidfowl 

i'm increasing the # of messages that load the first time you open a room so that people with larger screens get the scrollbar.

also, i'm turning up the # of messages to keep when the room is trimmed, just because.


davidfowl:
@samandmoore did you see the other bug related to trimming and loading history?

PR #776, opened 2013-04-03
Mobile Responsive Layouts - Inclusion of viewport and associated @media overrides

DavidChristiansen:
IMG_0706
IMG_0707

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -1026,15 +1026,13 @@
                 preferences.canToast = false;
                 $toast.hide();
             }
-            (function () {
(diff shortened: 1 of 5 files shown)


JabbR/Chat.ui.js
```diff
@@ -1026,7 +1026,15 @@
                 preferences.canToast = false;
                 $toast.hide();
             }
-
+            (function () {
+                if (navigator.userAgent.match(/IEMobile\/10\.0/)) {
+                    var msViewportStyle = document.createElement("style");
```
davidfowl: Why no jQuery?

davepermen: i guess because it's copypasted from the link i gave him about that fix, and, well, it works

PR #782, opened 2013-04-06
User list improvements

DavidChristiansen:
1. Replaced user status indicator bulb with greyed out username and pale grey background. Doing so increases space allowance for username, this means long names will abbreviated less.
   
2. Increased name width allowance on smallest screen size.


davidfowl:
We decided not to take these changes after discussion in meta.

PR #785, opened 2013-04-07
added editor config settings to force usage of spaces over tabs and ensu...

samandmoore:
...re 4 spaces.
- opt-in usage via http://editorconfig.org/ plugin

@davidfowl this will prevent me from having to constantly change my vs configuration, cool?


davidfowl:
Looks good

PR #790, opened 2013-04-08
Fix Google docs spreadsheet Content provider

jaircazarin:
The id of the document is now passed as part of the "key" query string
parameter instead of "formkey". Also there is a new URL for the embedded
view of a spreadsheet.


davidfowl:
Can you send this pull request to the dev branch?

Issue #806, opened 2013-04-11
Lobby filtering is broken


WilliamBZA:
![err](https://f.cloud.github.com/assets/1060960/366737/367fe922-a28f-11e2-8062-97777a885fd9.png)

Filtering by Meta doesn't filter properly.

davidfowl:
Fixed

Issue #787, opened 2013-04-07
"Load More..." doesn't work properly in lobby


cmaish:
If there are more than 100 rooms (or whatever limit is set), the lobby search will only operate on the rooms that are initially shown.

After clicking "Load More...", additional rooms are always shown, regardless of the search text provided.  The closed room filter seems to work properly against newly loaded elements.

If "Show Closed Rooms" is unchecked, "Load More..." will still display and appear to load no rooms (though I think that it's adding rooms that are currently hidden to the DOM)

PR #818, opened 2013-04-15
Room filtering

cmaish:
Several small changes to the lobby room list to:
- [x] address #806 
- [x] address #787
- [x] add empty list text
- [x] filter 'your private rooms' list
- [x] remove jquery livesearch
- [x] make closed private rooms showable
- [ ] make "Load more" less goofy when you have a filter (it might 'load' 100 new rooms but show none)
- [x] fix minor preexisting issues with ie9 and the icons (readonly / locked) not showing
- [ ] ensure that room creation / locking / update etc actually updates the various lists

What it looks like now (when no rooms match):
jabbr-rooms-filter


JabbR/Chat.ui.room.js
```diff
@@ -229,16 +229,14 @@
...
-                roomEmptyStatus.remove();
-                return;
-            }
+        var emptyStatus = list.children('li.empty'),
+            visibleItems = list.children('li:not(.empty):visible');
+        
+        if (visibleItems.length > 0) {
+            emptyStatus.remove();
+        } else if (emptyStatus.length === 0) {
+            list.append($('<li class="empty" />').text(list.data('empty-message')));
```
samandmoore, line 238: i think that call to data has to be `.data('emptyMessage')`

davidfowl:
We merged this change into dev and noticed that it completely broke on staging so the changes are now in https://github.com/JabbR/JabbR/commits/dev-fuuuu.

PR #820, opened 2013-04-15
old-ie gradients mess up rounding on connection status and user status

cmaish:
Unfortunately we lose some 3d-edness on these elements in IE9 and earlier.

The problem is described at http://stackoverflow.com/questions/6130235/ie9-rounded-corners-css-background-gradients-living-together and http://stackoverflow.com/questions/4692686/ie9-border-radius-and-background-gradient-bleeding

It's possible to work around with box-shadow (in the second link above) but it's a bit of a hack, and the graidents were quite subtle to begin with.

jabbr-ie9-status


hnrkndrssn:
Looks good to me :-D

PR #823, opened 2013-04-16
Room filtering

cmaish:
Actually make lobby room list work this time around.  There's still a bit to go around stuff like closing rooms (these aren't sorted at the moment) but should be functional.


davidfowl:
Wrong branch, send this to dev.

PR #827, opened 2013-04-17
Search using Lucene

samandmoore:
supports 2 providers:
- local filesystem (local to the web server)
- azure blob storage

The ui is SUPER barebones. We will want to get @alfhenrik to take a look at designing that :smile:


JabbR/Infrastructure/Lucene/LuceneIndexingJob.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public LuceneIndexingJob(TimeSpan interval, ISearchIndexingService searchIndexingService)
+            : base("SearchIndexing", interval)
+        {
+            _searchIndexingService = searchIndexingService;
+
+            // Updates the index synchronously first time job is created.
+            try
+            {
+                _searchIndexingService.UpdateIndex();
```
samandmoore: so this will synchronously create the index on app start, we probably don't want to do that. thoughts?

JabbR/Services/InMemoryRepository.cs
```diff
@@ -95,6 +95,16 @@ public void Remove(Notification notification)
...
+        public int GetMessageCountSince(int? lastMessageKey, out int newestMessageKey)
+        {
+            throw new NotImplementedException();
+        }
+
+        public IQueryable<ChatMessage> GetMessagesToIndex(int? lowerBoundKey, int upperBoundKey, int skip, int take)
+        {
+            throw new NotImplementedException();
+        }
+
```
samandmoore, line 127: doh, i need to implement these... there should probably be some unit tests against this stuff.

JabbR/App_Start/Startup.BackgroundWork.cs
```diff
@@ -34,10 +37,29 @@ private static void StartBackgroundWork(IKernel kernel, IDependencyResolver reso
             _sweepStart,
             _sweepInterval);
 
+            _jobManager = GetJobManager(kernel);
+            _jobManager.Start();
+
```
samandmoore, line 42: I think we should "stop" this at some point too, but I didn't know where the proper place to do that was with OWIN.

tkellogg:
You beat me to it. I was going to do this. Good think I checked first :)

tkellogg:
Also, I know from personal experience that Lucene isn't safe cross process. I'm not sure what the intended usage of Azure blob storage is, but don't do it if there's multiple web servers (cloud). There can only be one 'IndexWriter' per index. Instead, use Solr or elasticsearch

samandmoore:
@tkellogg thanks for the heads up. I think the objective for the cross-process access is to only ever write from one web-server. I think there is a web farm job coordinator in web backgrounder that we can use to handle that restriction (or we can probably set one up ourselves).

I would love to use solr or elasticsearch but then the issue we run into is, "how does an ordinary person download & build jabbr, then search?" Also, there is the cost factor ... I will see what @davidfowl thinks. We could probably put the abstraction in the right place and be able to support local filesystem search OR elasticsearch/solr via configuration settings.

Also, I'd love to talk to you about elasticsearch/solr if you have any experience using either one with .net. It would be nice to have something like that handle the incremental indexing for us, that is the biggest pain point in this impl.

tkellogg:
@samandmoore Unfortunately, I don't have much experience with Solr or elasticsearch. At work, my CTO is of the "reinvent the wheel" school of thought (self-proclaimed, seriously...), so we wrote our own crappy version of Solr. We used a database table as a message queue. Write operations are queued up and performed by a single web node, while all read operations skip the queue and happen synchronously. Writing the whole thing was tedious enough that I have a lot of respect for what Solr & elasticsearch do. (FYI they're both just cloud-based REST APIs for working with a Lucene index). I wish I could throw out my own code and just use one of them.

Honestly, I'd highly recommend against the typical deployment using something like Solr or elasticsearch. Perhaps the jabbr.net site might need that sort of scalability, but I know I don't need it (for my team's instance, running off of my desktop). Having the default implementation be a Lucene FSDirectory makes a lot of sense. As long as you have an abstraction _over top_ Lucene, you should be able to swap it out for elasticsearch or some other cloud-based search service.

davidfowl:
@samandmoore Don't merge this in as yet as I want to finish reviewing it and I also want to merge dev into master once it's stable.

davidfowl:
@samandmoore show me a screenshot

JabbR/App_Start/Startup.BackgroundWork.cs
```diff
@@ -34,10 +37,29 @@ private static void StartBackgroundWork(IKernel kernel, IDependencyResolver reso
...
             // Clear all connections on app start
             ClearConnectedClients(kernel);
         }
 
+        private static JobManager GetJobManager(IKernel kernel)
```
davidfowl, line 47: What's this? Where is the JobManager from? WebBackgrounder?

JabbR/App_Start/Startup.DependencyInjection.cs
```diff
@@ -121,6 +121,26 @@ private static KernelBase SetupNinject(ApplicationSettings settings)
...
+            else
+            {
+                luceneFileSystem = new LuceneAzureBlobFileSystem(settings);
+            }
+
+            kernel.Bind<ILuceneFileSystem>()
+                  .ToConstant(luceneFileSystem);
+
+            kernel.Bind<ISearchIndexingService>()
+                  .ToMethod(context => new LuceneIndexingService(() => new PersistedRepository(new JabbrContext()), context.Kernel.Get<ILuceneFileSystem>()))
```
davidfowl, line 138: Shouldn't you resolve the IJabbRRepository from the kernel?

JabbR/Infrastructure/Lucene/LuceneAzureBlobFileSystem.cs
```diff
@@ -0,0 +1,134 @@
...
+        }
+
+        private AzureDirectory _indexDir;
+        public Directory IndexDirectory
+        {
+            get
+            {
+                if (_indexDir == null)
+                {
+                    _indexDir = new AzureDirectory(_account, JabbRSearchIndexContainer, new RAMDirectory());
```
davidfowl, line 34: Any good reason this is lazy?

JabbR/Infrastructure/Lucene/LuceneIndexingJob.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        public LuceneIndexingJob(TimeSpan interval, ISearchIndexingService searchIndexingService)
+            : base("SearchIndexing", interval)
+        {
+            _searchIndexingService = searchIndexingService;
+        }
+
+        public override Task Execute()
+        {
+            return new Task(_searchIndexingService.UpdateIndex);
```
davidfowl, line 19: New Task? Is that the pattern of WebBackgrounder? Cold tasks?

JabbR/Infrastructure/Lucene/LuceneIndexingService.cs
```diff
@@ -0,0 +1,190 @@
...
+                return new IndexTaskMetaData()
+                {
+                    LowerBoundMessageKey = lastMessageKey,
+                    UpperBoundMessageKey = newestMessageKey,
+                    TotalItemsToIndex = messageCountToIndex,
+                };
+            }
+        }
+
+        private void AddChatMessages(IndexTaskMetaData indexTaskMetaData)
```
davidfowl, line 77: So we're building the index 100 messages at a time right? We need to option to run this same logic offline (when the index is first built). Or somehow split it up over multiple jobs.

samandmoore, line 77: I will try to take a look at this tomorrow. Sorry for the delay! I have
been really busy lately.

JabbR/App_Start/Startup.BackgroundWork.cs
samandmoore, line 47: yeah that's from web backgrounder.

JabbR/Infrastructure/Lucene/LuceneAzureBlobFileSystem.cs
samandmoore, line 34: not really, I was trying to avoid any expensive costs of the AzureDirectory ctor. Since I think it does some connection work.

JabbR/Infrastructure/Lucene/LuceneIndexingJob.cs
samandmoore, line 19: Yeah, that appears to be the pattern. You return a task object that hasn't been started, and it will start it up.

samandmoore:
closing this for now, it needs a lot of thought.

PR #834, opened 2013-04-25
use insertion sort for user list sorting and removed .filter call for em...

samandmoore:
...pty lists in favor of pseudo selector

might help with the slowness with huge user lists


davidfowl:
I think the problem isn't the sorting itself but the DOM redraw after we append.

PR #836, opened 2013-04-26
Hide user list container element before sorting

### LINKED ISSUE #835: DOM reflow issues when updating user lists
During the DotNetConf today I noticed massive slowdown and, sometimes, outright hanging of IE as the room list grew. I profiled for a little bit and it appears that the sorting of the room user list ultimately causes O(N) calls to append which are forcing O(N) DOM reflows.

Here's a profiling screenshot where you can see how each call to addUser is ultimately calling addUserToList which ultimately calls sortList each time. If you drill in further (not shown in this screenshot) you'll see that the majority of the time there is attributable to the O(N) append calls that are occurring inside of sortList.

JabbRAddUser_JavaScriptProfileTree

A simple fix may just be to set the user list container element to display:none; prior to sorting. That could be potentially jarring depending on several layout factors, but worth a try in the short term.

If the display:none trick doesn't work, the child elements may need to be moved into a documentFragment before sorting and then moved back out into the live DOM after sorting has been completed.

drub0y:
This attempts to fix #835 by hiding the user list container element using display:none; and then restoring it to its original display value after the sorting of the child nodes is completed.

I also added an explicit length test to see if there were any children even in a particular list (owners vs. activeusers) as a short circuit before doing any more costly operations unecessarily.

### DIFF
--- a/JabbR/Chat.ui.room.js
+++ b/JabbR/Chat.ui.room.js
@@ -315,26 +315,33 @@
     Room.prototype.sortList = function (listToSort) {
         var listItems = listToSort.children('li:not(.empty)').get();
 
-        var activeUsers = [],
-            idleUsers = [],
-            sortedUsers = [];
+        if (listItems.length > 0) {
+            var originalDisplayMode = listToSort.css("display"),
+                activeUsers = [],
+                idleUsers = [],
+                sortedUsers = [];
 
-        $.each(listItems, function (index, item) {
-            if ($(item).data('active')) {
-                activeUsers.push(item);
-            } else {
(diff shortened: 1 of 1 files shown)


davidfowl:
Does it work? We can test this easily by faking up data and verifying new adds and removal of users. If it's too jarring I'm not willing to take it, even if it is temporary.

drub0y:
I'm closing this because it will not sufficiently address the issue. The only way to solve this problem is to stop O(N)'ing the list M times as the code currently does which is going to require much bigger changes.

PR #837, opened 2013-05-01
Add ability to set TraceWriter/TraceLevel on connection before starting

cwensley:


JabbR.Client/JabbRClient.cs
```diff
@@ -102,6 +105,28 @@ public ICredentials Credentials
             }
         }
 
+        public TextWriter TraceWriter
+        {
+            get { return _connection != null ? _connection.TraceWriter : _traceWriter; }
```
davidfowl: Why not just use auto properties?

cwensley: So one can set the tracelevel/writer after the connection is created

davidfowl: I dont think it matters anyway

cwensley: I suppose.. most of jabbr.client needs to be setup before connecting (events, etc).  Not that I like that it does that.. (;

PR #841, opened 2013-05-07
Update the lock/readonly icons to use FontAwesome for scaling

### LINKED ISSUE #840: Padlock Icon needs updating
The padlock icon is old and doesn't scale for Retina display. So all the icons look crystal clear except that one.

Need to replace it with a nice image that scales properly.

hnrkndrssn:
I had to update the Closed/ReadOnly icon as there isn't a crossed out chat bubble icon in the FontAwesome arsenal, but I believe the ban circle icon conveys the message...let me know if you'd like me to change it :smiley: 

Zoomed to 200% on Non-Retina
image

Zoomed to 150% on Non-Retina
image

Normal Non-Retina
image

Fixes #840

### DIFF
--- a/JabbR/Chat.css
+++ b/JabbR/Chat.css
@@ -481,17 +481,15 @@ form#send-message {
     }
 
     .users li.room.locked .lock {
-        position: absolute;
-        width: 16px;
+        position: relative;
+        width: 19px;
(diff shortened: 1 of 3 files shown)


davidfowl:
Looks good.

PR #851, opened 2013-05-17
Add allowed command

### LINKED ISSUE #849: Add command to list the users allowed in a private room
You can add/delete users today, but apparently there is no way to list them.

cmaish:
Should fix #849

There are a couple of bits that aren't great - exposing EnsureAllowed on IChatService is probably the worst, and I dunno how good the tests are (they do pass though).

Lets you type /allowed [roomname] where roomname is optional (defaults to the current room).  If you could join the room, it'll list the users who are allowed to join the room (or tell you "Anyone is allowed in general-chat as it is not private" if it's public).

### DIFF
--- a/JabbR.Tests/CommandManagerFacts.cs
+++ b/JabbR.Tests/CommandManagerFacts.cs
@@ -3362,7 +3362,6 @@ namespace JabbR.Test
                     Name = "room",
                     Private = true,
                     AllowedUsers = new Collection<ChatUser>() { user },
-                    Users = new Collection<ChatUser>() { user }
                 };
                 repository.Add(room);
                 var room2 = new ChatRoom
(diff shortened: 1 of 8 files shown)


JabbR.Tests/CommandManagerFacts.cs
```diff
@@ -10,6 +10,8 @@
 
 namespace JabbR.Test
 {
+    using System.Collections.ObjectModel;
```
davidfowl, line 13: using inside the namespace?

JabbR/Commands/AllowedCommand.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace JabbR.Commands
+{
+    [Command("allowed", "Show a list of all users allowed in the given room.", "[room]", "room")]
+    public class AllowedCommand : UserCommand
+    {
+        public override void Execute(CommandContext context, CallerContext callerContext, ChatUser callingUser, string[] args)
+        {
+            string targetRoomName = args.Length > 0 ? args[0] : callerContext.RoomName;
+
+            if (string.IsNullOrEmpty(targetRoomName) || targetRoomName.Equals("Lobby", StringComparison.InvariantCultureIgnoreCase))
```
davidfowl, line 13: Uppercase S

Issue #852, opened 2013-05-20
Unallow user causes exception sometimes


cmaish:
Unallow user sometimes causes an exception:

"An error occurred while saving entities that do not expose foreign key properties for their relationships. The EntityEntries property will return null because a single entity cannot be identified as the source of the exception. Handling of exceptions while saving can be made easier by exposing foreign key properties in your entity types. See the InnerException for details."

This seems to be related to trying to remove users from the room if they're not currently present, and seems to be intermittent.

cmaish:
Dupe of #854

PR #853, opened 2013-05-20
Prevent trying to remove users from the room if they're not already present

cmaish:
May help mitigate #852, guard against users that aren't present being removed from the list, possibly causing an exception.

### DIFF
--- a/JabbR/Services/PersistedRepository.cs
+++ b/JabbR/Services/PersistedRepository.cs
@@ -196,6 +196,11 @@ namespace JabbR.Services
         {
             RunNonLazy(() =>
             {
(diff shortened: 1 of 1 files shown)


JabbR/Services/PersistedRepository.cs
```diff
@@ -196,6 +196,11 @@ public void RemoveUserRoom(ChatUser user, ChatRoom room)
         {
             RunNonLazy(() =>
             {
+                if (!room.Users.Contains(user))
```
davidfowl, line 199: That's no good as room.Users is what we're trying to avoid in the first place since room.Users loads all users

PR #857, opened 2013-05-21
Improve room commands

cmaish:
[take 2, this time against dev branch]

Make room commands take the room name optionally (default to current room), and handle access slightly more gracefully - using "is the user allowed into the room or owner or admin" rather than "is the user currently in the room" to account for this.

### DIFF
--- a/JabbR/Hubs/Chat.cs
+++ b/JabbR/Hubs/Chat.cs
@@ -823,9 +823,10 @@ namespace JabbR
 
         void INotificationService.PostNotification(ChatRoom room, ChatUser user, string message)
         {
+            string roomName = room == null ? null : room.Name;
             foreach (var client in user.ConnectedClients)
             {
-                Clients.Client(client.Id).postNotification(message, room.Name);
+                Clients.Client(client.Id).postNotification(message, roomName);
             }
         }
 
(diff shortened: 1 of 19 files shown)


JabbR/Hubs/Chat.cs
```diff
@@ -823,9 +823,10 @@ void INotificationService.SendPrivateMessage(ChatUser fromUser, ChatUser toUser,
 
         void INotificationService.PostNotification(ChatRoom room, ChatUser user, string message)
         {
+            string roomName = room == null ? null : room.Name;
```
davidfowl: Why would it ever be null?

cmaish: Invitecode and resetinvitecode both had to be changed to try to post to whatever room you're in (previously they posted to the room that you were getting the code for).  If you're not in a room (at the moment this means in the lobby), this will wind up being null, and this change allows responses to be sent to the lobby's log (even though in the web client this log is currently hidden).  I figure that it was better than throwing an exception when you try to perform certain commands from outside a room.

davidfowl: I dunno how I feel about that.

cmaish: Only two commands tried to use that functionality.  I've updated the commands to throw an invalid operation exception if you try to run them from the lobby and reverted the roomName == null stuff.

PR #861, opened 2013-05-26
Limit middle click tab header action to tabs for rooms

### LINKED ISSUE #860: Middle mouse click on lobby tab header shows error message if you're in a room
When you middle click on a tab header for a room, the tab is closed.  The lobby is unclosable and isn't a room.

If you're in a room (ie somewhere other than the lobby) and middle click on the lobby, the operation will fail with an error message of "Unable to find room 'Lobby'".  

The bug is the error message text not being appropriate.  I think that it should not show an error (and middle clicks on the lobby tab header should be ignored).

cmaish:
Fix #860 by not attempting to close non-room tabs when you middle-mouse-click on them.

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -839,7 +839,7 @@
                 ui.setActiveRoom($(this).data('name'));
             });
 
-            $document.on('mousedown', '#tabs li', function (ev) {
+            $document.on('mousedown', '#tabs li.room', function (ev) {
                 // if middle mouse
                 if (ev.which === 2) {
                     $ui.trigger(ui.events.closeRoom, [$(this).data('name')]);


samandmoore:
this failed jshint, details:

Chat.ui.js: line 778, col 56, Expected an assignment or function call and instead saw an expression.

---

EDIT, this was caused by a different PR, i fixed it.

JabbR/Chat.ui.js
```diff
@@ -839,7 +839,7 @@
                 ui.setActiveRoom($(this).data('name'));
             });
 
-            $document.on('mousedown', '#tabs li', function (ev) {
+            $document.on('mousedown', '#tabs li.room', function (ev) {
```
samandmoore, line 842: did you consider checking something like:

`if (room.isLobby()) return false`?

cmaish, line 842: I looked at doing the change that way before I settled on this way - it's definitely possible (and straightforward), but I went for the simpler / smaller change.  The thing that particularly pushed me away from that method was that getting the room constructs a Room object / running a bunch of jquery selectors that won't be used for anything.

I'm happy to change the PR to do this (getting the data-name attribute from the tab, constructing the room, testing whether it matches the lobby) if this one fails for readability reasons though of course.

davidfowl, line 842: What's less risky?

cmaish, line 842: Neither change is particularly risky/complex.  What I did is less code and results in less stuff happening during execution (the selector is skipped completely by jquery if it's not a room if you go this path; the other way we'd write logic that skips it ourselves).

davidfowl, line 842: Well the question really is about breaking things that used to work. The way I see it, if there was something else without the room class that we're overlooking, your code will break it. If you're certain that that isn't the case then I prefer your fix, otherwise it's less risky to filter ourselves.

cmaish, line 842: That particular event handler was written by me pretty recently (https://github.com/cmaish/JabbR/commit/4c7b1f8129e931566f7a2b62bc97dc6e7069731e#JabbR/Chat.ui.js) and I made it too broad.  The entire purpose of it was to try to close whatever tab you middle click on, and while I tested that it didn't close the lobby, I didn't notice that it gave an error message if you try to close the lobby while you're focussed in another room.

I'm quite confident that in the current codebase this will work and have the desired effect.  The only reason that I can think of that we might want a more broad handler is if we introduce non-lobby, non-room tabs that we want to be able to close, though these are going to require their own code anyways (as the closeRoom event which is raised here won't be appropriate, etc).

davidfowl, line 842: Ok you sold me.

PR #863, opened 2013-05-26
Use full username in message template

### LINKED ISSUE #862: Long usernames in message history handle clicks incorrectly
When you click on a username from either the users menu or a username to the left of a message, the message textarea is supposed to prepend "@UserNameYouClicked" to allow you to write an @ reply.

If a user's name is more than around 21 characters, clicking on the users menu works, but clicking on the username to the left of a message gives something like "@ARelativelyLongNam..."

The underlying cause for this is that a truncated version of the name was added to the templates in the past (to allow for long names without messing up the display of the chat text).  A subsequent change hid overflowing text (and added ellipses) via CSS, so using this shortened username is no longer necessary.

cmaish:
Fixes #862 . Truncation isn't necessary because the CSS handles overflow (and adds ellipses)

### DIFF
--- a/JabbR/Views/Home/index.cshtml
+++ b/JabbR/Views/Home/index.cshtml
@@ -46,7 +46,7 @@
             <div class="left">
                 {{if showUser}}
                 <img src="https://secure.gravatar.com/avatar/${hash}?s=16&d=mm" class="gravatar" />
-                <div class="name">${trimmedName}</div>
+                <div class="name">${name}</div>
                 {{/if}}
                 <span class="state"></span>
             </div>
--- a/JabbR/Chat.toast.js
+++ b/JabbR/Chat.toast.js
@@ -24,10 +24,10 @@
                 return;
             }
 
-            var toastTitle = message.trimmedName;
+            var toastTitle = utility.trim(message.name, 21);
             // we can reliably show 22 chars
             if (toastTitle.length <= 19) {
-                toastTitle += ' (' + utility.trim(roomName, 19 - message.trimmedName.length) + ')';
+                toastTitle += ' (' + utility.trim(roomName, 19 - toastTitle.length) + ')';
             }
 
             toastRoom = roomName;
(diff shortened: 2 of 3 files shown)


JabbR/Views/Home/index.cshtml
```diff
@@ -46,7 +46,7 @@
             <div class="left">
                 {{if showUser}}
                 <img src="https://secure.gravatar.com/avatar/${hash}?s=16&d=mm" class="gravatar" />
-                <div class="name">${trimmedName}</div>
+                <div class="name">${name}</div>
```
samandmoore, line 49: didn't we trim the name for a reason? I think it would be better to leave this but then add a data attribute with the full name, then use that to fix the related issue.

samandmoore, line 49: doh, just saw your comment. does that css bit for adding the ellipsis work in all modern browsers?

cmaish, line 49: As far as I can tell, text-overflow is pretty well supported:

http://www.w3schools.com/cssref/css3_pr_text-overflow.asp
https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

It was vendor prefixed in IE6.  If it's not available, the name will be shown, truncated without ellipsis due to the overflow:hidden that's also specified on the .message .left .name class, but I wouldn't expect that fallback to come into play in any browser that can actually run jabbr.

samandmoore, line 49: okay then that's good for me.

samandmoore, line 49: are we still using trimmedName anywhere? it should be removed if not.

cmaish, line 49: trimmedName was being used in the toast header, but the logic should probably be tweaked at some point (and is specific to how the titles on toasts display), so I inlined it there and removed it from the message object.

PR #864, opened 2013-05-26
Prevent ctrl-tab from switching jabbr tab in Firefox

### LINKED ISSUE #722: Ctrl-Tab switches JabbR tabs in Firefox
When using Ctrl-Tab to switch browser-tabs while using firefox, Navigation within JabbR switches to the next JabbR Tab, too. Apparently, Ctrl-Tab gets interpreted as Tab in firefox.

cmaish:
In Firefox this event gets sent to javascript as well as handled by the browser, so when a user does this in FF they usually come back to a different room than they left.

Fixes #722

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -904,7 +904,8 @@
             // handle tab cycling - we skip the lobby when cycling
             // handle shift+/ - display help command
             $document.on('keydown', function (ev) {
-                if (ev.keyCode === Keys.Tab && $newMessage.val() === "") {
+                // ctrl + tab event is sent to the page in firefox when the user probably means to change browser tabs
+                if (ev.keyCode === Keys.Tab && !ev.ctrlKey && $newMessage.val() === "") {
                     var current = getCurrentRoomElements(),
                         index = current.tab.index(),
                         tabCount = $tabs.children().length - 1;


JabbR/Chat.ui.js
```diff
@@ -904,7 +904,8 @@
             // handle tab cycling - we skip the lobby when cycling
             // handle shift+/ - display help command
             $document.on('keydown', function (ev) {
-                if (ev.keyCode === Keys.Tab && $newMessage.val() === "") {
+                // ctrl + tab event is sent to the page in firefox when the user probably means to change browser tabs
+                if (ev.keyCode === Keys.Tab && !ev.ctrlKey && $newMessage.val() === "") {
```
samandmoore, line 908: this is cross-browser right? I'm just thinking of IE and ev.which stuff.

cmaish, line 908: ctrlKey is in the DOM level 2 spec (http://www.w3.org/TR/2000/REC-DOM-Level-2-Events-20001113/events.html#Events-MouseEvent-ctrlKey), and AFAIK is supported by every browser since Netscape Navigator 4.  Regardless, because of how we're filtering it out, if the property didn't exist in some modern browser, ev.ctrlKey would be undefined, which would enter the tab-switching routine (so the system would still behave as expected).

samandmoore, line 908: thanks for the reference. looks good to me.

Issue #867, opened 2013-06-01
Username change often messes up userlist and message history.


cmaish:
When you call changeUserName, the users messages in the room seem to be sorted along with the username itself.

We want to update the username references on all messages, but only resort the user item itself rather than all messages they've posted.

To repro, have 2 regular users in a room, with some chats from each.  Rename one of the users.  The user list might look like: https://jabbr.blob.core.windows.net/jabbr-uploads/clipboard_e370349e46cf4d048ab221f9ab79eec1_ad77.png

PR #868, opened 2013-06-01
Make username change not break things

cmaish:
Currently, when you change your username, any rooms that you're in get broken for everyone if you've got messages - #867.  This is due to sorting the messages from the user into the userlist.  This change fixes that.


hnrkndrssn:
Looks good to me

PR #871, opened 2013-06-04
Added Screencast.com content provider

Redth:
This adds the ability to show images from Screencast.com (often uploaded with a tool called snagit) as a content provider.

An example url/image: http://www.screencast.com/t/VzUgRVW9yMr8

### DIFF
--- a/JabbR/ContentProviders/ScreencastContentProvider.cs
+++ b/JabbR/ContentProviders/ScreencastContentProvider.cs
@@ -30,7 +30,9 @@ namespace JabbR.ContentProviders
 
         private Task<PageInfo> ExtractFromResponse(ContentProviderHttpRequest request)
         {
-            return Http.GetAsync(request.RequestUri).Then(response =>
+            var builder = new UriBuilder(request.RequestUri) { Scheme = "https" };
+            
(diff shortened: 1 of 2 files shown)


davidfowl:
Is there an https version? I'm trying to avoid any new content providers that don't support https as it causes that horrible warning to appear.

Redth:
Looks like https works.. let me change the code to force https :)

JabbR/ContentProviders/ScreencastContentProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        private static readonly string ContentFormat = "<img src=\"{0}\" alt=\"{1}\" />";
+
+        protected override Task<ContentProviderResult> GetCollapsibleContent(ContentProviderHttpRequest request)
+        {
+            return ExtractFromResponse(request).Then(pageInfo =>
+            {
+                return new ContentProviderResult
+                {
+                    Content = String.Format(ContentFormat, pageInfo.ImageURL, pageInfo.Title),
```
davidfowl: You need to HtmlAttribute encode those properties.

PR #878, opened 2013-06-05
Improve consistency of behavior around closed rooms

cmaish:
At current, closed rooms allow you to enter them with an invite code, however allow, invite, lock, unallow do now work for closed rooms.  This PR makes those base-level user-control functions available.

Note that things that affect users IN the room are unavailable.  Owners cannot be added or removed.

The big part of this change is the change to IJabbrRepository.VerifyUserInRoom to add the optional mustBeOpen parameter.  This is called by the me, nudge, topic, welcome commands, as well as from the chat hub and chatservice.  For the service and hub, the call sites have been updated to include mustBeOpen: false.  For the four commands, previously users could set those (even if it wasn't possible without messing with the UI), now they cannot.

I've added a bunch of tests to exercise this functionality.

### DIFF
--- a/JabbR/Hubs/Chat.cs
+++ b/JabbR/Hubs/Chat.cs
@@ -159,15 +159,14 @@ namespace JabbR
             var userId = Context.User.GetUserId();
 
             ChatUser user = _repository.VerifyUserId(userId);
-            ChatRoom room = _repository.VerifyUserRoom(_cache, user, clientMessage.Room);
+            ChatRoom room = _repository.VerifyUserRoom(_cache, user, clientMessage.Room, mustBeOpen: false);
 
             if (room == null || (room.Private && !user.AllowedRooms.Contains(room)))
             {
                 return false;
             }
 
-            // REVIEW: Is it better to use _repository.VerifyRoom(message.Room, mustBeOpen: false)
-            // here?
+            // REVIEW: Is it better to use _repository.VerifyUserRoom(..., mustBeOpen: true) here?
             if (room.Closed)
             {
                 throw new InvalidOperationException(String.Format("You cannot post messages to '{0}'. The room is closed.", clientMessage.Room));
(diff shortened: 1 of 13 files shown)


JabbR/Hubs/Chat.cs
```diff
@@ -159,15 +159,14 @@ public bool Send(ClientMessage clientMessage)
             var userId = Context.User.GetUserId();
 
             ChatUser user = _repository.VerifyUserId(userId);
-            ChatRoom room = _repository.VerifyUserRoom(_cache, user, clientMessage.Room);
+            ChatRoom room = _repository.VerifyUserRoom(_cache, user, clientMessage.Room, mustBeOpen: false);
```
davidfowl: Uh wut?

JabbR/Hubs/Chat.cs
```diff
@@ -471,7 +470,7 @@ public void Typing(string roomName)
             string userId = Context.User.GetUserId();
 
             ChatUser user = _repository.GetUserById(userId);
-            ChatRoom room = _repository.VerifyUserRoom(_cache, user, roomName);
+            ChatRoom room = _repository.VerifyUserRoom(_cache, user, roomName, mustBeOpen: false);
```
davidfowl: Whut?

JabbR/Hubs/Chat.cs
cmaish: In almost every case that we use VerifyUserRoom, we want to both verify that the user is in the room and that the room that the user is in is open.  As a result, I altered the VerifyUserRoom to take a mustBeOpen parameter like VerifyRoom, where it can throw an InvalidOperationException when the room isn't open.

I made this change to keep things more DRY so that I didn't have to keep repeating the exception in each command that verified that the user was in a room (but cared that the room that the user was in wasn't open anymore).

It seemed neater to do this than to call VerifyRoom immediately after or before VerifyUserRoom in the commands, because they have some other overlap in functionality.  It might be more clear if the parameter was renamed roomMustBeOpen, but that won't solve any concerns with the actual structure of the change.

PR #883, opened 2013-06-12
Improve consistency of InvalidOperationException messages

cmaish:
Currently, the wording, length and punctuation on InvalidOperationExceptions (which are shown to the user when something fails in a command) are inconsistent.

This change changes the messages to, where appropriate, use phrasing like:

"Who do you want to (perform action phrase here)?"
"Which room do you want to (perform action phrase here)?"

and ensures punctuation at the end of each message.  I've also extended the tests to verify the messages that are returned so that we're more aware if regressions occur.


davidfowl:
While you're at it do you want to move the exceptions to resources so jabbr can be localized :smile:

PR #889, opened 2013-06-16
First steps to localisation

cmaish:
Created a hueg resource file that contains pretty much all of the strings that should benefit from localisation.
- [x] Replace exception messages with strings from the resources.
- [x] Update tests where relevent.
- [x] Replace help message descriptions with resource-based versions.
- [x] Update templates to use resource-based strings.
- [x] Dump localisation object containing strings used by javascript in bottom of main index template.
- [x] Make javascript actually use these.
- [x] Make messages more consistent in terms of punctuation, tense, wording.
- [x] Remove no-longer-relevant references to nick or nickname (now "user"), update command info to capture current system (stuff like notes aren't shown with a paper-clip anymore)
- [x] Bring Nancy culture to the thread.
- [x] Make Nancy specify UTF-8.
- [x] Make sure everything is bulletproof. (Partially done).
- [x] RFC on changes in general.
- [x] Figure out if it's worth (or possible to) allow in-client localisation or having a separate app for localisation.  (nah, it's not worth either at this stage if it's going to take me another week to get around to looking at it).
- [ ] Add a user option to allow overriding the language sent by the browser.

There may be a couple of rough edges and things that can benefit from being refactored, but all of the major pieces should be there and done.


JabbR.Tests/CommandManagerFacts.cs
```diff
@@ -153,7 +153,7 @@ public void ThrowsIfCommandDoesntExist ()
                 var commandManager = new CommandManager("1", "1", "room", service, repository, cache, notificationService.Object);
 
                 InvalidOperationException ex = Assert.Throws<InvalidOperationException>(() => commandManager.TryHandleCommand("/message"));
-                Assert.Equal("'message' is not a valid command.", ex.Message);
```
davidfowl, line 156: Why'd you remove the quotes?

JabbR.Tests/CommandManagerFacts.cs
```diff
@@ -3041,38 +3041,13 @@ public void ClosedRoomThrows()
...
                 InvalidOperationException ex = Assert.Throws<InvalidOperationException>(() => commandManager.TryHandleCommand("/me is testing"));
-                Assert.Equal("The room 'room' is closed.", ex.Message);
+                Assert.Equal("room is closed.", ex.Message);
             }
         }
 
         public class MsgCommand
         {
             [Fact]
-            public void ThrowsIfThereIsOnlyOneUser()
```
davidfowl, line 3051: Why did this change?

JabbR/Chat.js
```diff
@@ -248,7 +248,7 @@
 
         if (added) {
             populateRoom(room.Name).done(function () {
-                ui.addMessage('You just entered ' + room.Name, 'notification', room.Name);
+                ui.addMessage(String.formatLanguageResource('Chat_YouEnteredRoom', room.Name), 'notification', room.Name);
```
davidfowl: lowercase format? What method is this?

davidfowl:
I think I like the approach in general. Does this allow clients to choose different languages or is it a single version of jabbr where all clients and the server are in a single language?

JabbR.Tests/CommandManagerFacts.cs
cmaish, line 156: The quotes were used inconsistently through other messages.  For things like welcome, room topic and messages which can contain spaces it might make sense for there to be quotes.  Elsewhere there was a mix of "you have been xyzed from room 'general-chat'", "you have been xyzed from general-chat" and "you have been xyzed from 'general-chat'" so I took out quotes and the word 'room'.  I don't mind if you have a preference either way - adding that stuff back should be a simple case of updating the resx strings now.

JabbR.Tests/CommandManagerFacts.cs
cmaish, line 3051: It was redundant (and one less string to localise if taken out) - if you were the only user known to the entire jabbr instance, we used to get that specific error message.  It's helpful to know that there's not anyone else who you can ever message, but the only person who's going to ever see it is an admin, and other messages tell the user they can't message themselves.

JabbR/Chat.js
cmaish: A new one in the utilities javascript.  There's also a format method so that we can use the same format strings that we use for server-side localisation stuff.

cmaish:
Per-user/client, based on the basic process described at http://www.jefclaes.be/2013/04/internationalization-in-nancyfx.html.  IE's Internet Options -> Languages allows you to specify what gets sent up and is probably the easiest way to check things if you add another language.

We might want to bump the cookie detection up the order and do something like pulling culture from the flag, or add another culture field on the user properties screen or something.

JabbR/Chat.js
davidfowl: Change this to getLanguageResource and move it off String.

JabbR/Views/Home/index.cshtml
```diff
@@ -453,7 +481,94 @@
     }
     <script type="text/javascript" src="@Url.Content("~/signalr/hubs")"></script>
     <script type="text/javascript">
-        window.jabbrVersion = '@Model.Version';
+      window.jabbrVersion = '@Model.Version';
+      window.languageResources = {
```
davidfowl, line 471: Move this into C#.

JabbR/Views/Home/index.cshtml
```diff
@@ -384,25 +405,34 @@
                         {
                             <div id="releaseArea">
                                 <p id="releaseTag">
-                                    Deployed from <a target="_blank" href="https://github.com/JabbR/JabbR/commit/@Model.Sha" title="View the commit.">@Model.Sha.Substring(0, 10)</a>
-                                    on <a target="_blank" href="https://github.com/JabbR/JabbR/branches/@Model.Branch" title="View the branch.">@Model.Branch</a> 
-                                    at @Model.Time.
+                                    @{{
```
davidfowl: Don't like this conversion. Just do inline rendering.

davidfowl:
Failing travis ci

JabbR/Views/Home/index.cshtml
```diff
@@ -384,25 +396,26 @@
                         {
                             <div id="releaseArea">
                                 <p id="releaseTag">
-                                    Deployed from <a target="_blank" href="https://github.com/JabbR/JabbR/commit/@Model.Sha" title="View the commit.">@Model.Sha.Substring(0, 10)</a>
-                                    on <a target="_blank" href="https://github.com/JabbR/JabbR/branches/@Model.Branch" title="View the branch.">@Model.Branch</a> 
-                                    at @Model.Time.
+                                    @Html.Raw(String.Format(Encoder.HtmlEncode(LanguageResources.Client_DeploymentInfo),
```
davidfowl: Same here, render it with razor.

cmaish:
I'm not a huge fan of https://github.com/cmaish/JabbR/commit/3e519196121852e971438f7c35a7ea5b1a0daaed in general.  It isn't DRY, markup exists in resources, and HtmlSanitizationLibrary is broken to the point of being unusable (it strips href tags out of anchors) so we need to trust that resource files aren't malicious.

JabbR/Nancy/JabbRNinjectNancyBootstrapper.cs
```diff
@@ -58,6 +64,21 @@ private Response FlowPrincipal(NancyContext context)
...
+            Thread.CurrentThread.CurrentCulture = ctx.Culture;
+            Thread.CurrentThread.CurrentUICulture = ctx.Culture;
+            return null;
+        }
+
+        private void SetContentType(NancyContext ctx)
+        {
+            if (ctx.Response.ContentType.Contains("text/html"))
+            {
+                ctx.Response.ContentType = "text/html; charset=utf-8";
```
davidfowl: Can we always do this? (append utf8)

cmaish: There's probably a difference in "can" vs. "should" here, but I couldn't find a huge amount of info on the issue itself.  The most relevant discussion is https://groups.google.com/forum/#!msg/nancy-web-framework/07G3oQF8XAM/Nkrn4FYRudMJ which resulted in a changeset that explicitly sets all of the response streams to UTF8 with Byte Order Marks in one branch - though this didn't get pushed back to the main Nancy project for some reason (https://github.com/kleinron/Nancy/commit/01808a1ae3894b165f64bf555fc5102272e3e98b).

I can't say for certain, and for that matter I'm not certain why Nancy leaves detection up to the browser, but it definitely seems safe to include the encoding (and necessary when using non-latin characters).

JabbR/Services/UploadCallbackHandler.cs
```diff
@@ -71,7 +73,7 @@ public class UploadCallbackHandler
         private static string FormatBytes(long bytes)
         {
             const int scale = 1024;
-            string[] orders = new string[] { "GB", "MB", "KB", "Bytes" };
+            string[] orders = new string[] { LanguageResources.SizeOrderGB, LanguageResources.SizeOrderMB, LanguageResources.SizeOrderKB, LanguageResources.SizeOrderBytes };
```
davidfowl, line 76: Does this really need to be localized?

cmaish, line 76: It's always the French, the correct way to write these in French is 10 Go, 10 Mo, 10 Ko etc

davidfowl:
@cmaish update the pr and I'll merge it in

PR #891, opened 2013-06-17
Updated flags from flag-sprites.com

damieng:


davidfowl:
@damieng No flags were removed right?

damieng:
It doesn't look like it but some CSS names for things that don't have ISO-3166 country codes seem to have changed,.

Changes:
g1 -> england
g2 -> scotland
g2 -> wales

Added:
cw
ic 
kurdistan
somaliland
zanzibar

PR #895, opened 2013-06-18
Add new country codes to Jabbr.

damieng:
This patch adds the missing country codes from flag-sprites.png.
As there are additional non-ISO-3166 codes some are not 2 characters in length so DB is extended.
Old G1/G2/G3 codes now mapped to England/Wales/Scotland.


JabbR/Services/ChatService.cs
```diff
@@ -857,7 +869,14 @@ internal static string GetCountry(string isoCode)
             }
 
             string country;
-            return Countries.TryGetValue(isoCode, out country) ? country : null;
+            if (Countries.TryGetValue(isoCode, out country))
```
davidfowl, line 872: Add the braces, no self closing ifs.

PR #901, opened 2013-06-20
Fixes JabbR/JabbR#874

johnbfair:


davidfowl:
This change broke all of the images references in css which I fixed in https://github.com/JabbR/JabbR/commit/82f4b3554d670b205bb617262b56bc85dee0262e just an FYI.

Issue #904, opened 2013-06-24
Preventative measures against bots


PureKrome:
One damn bot is being a damn PITA.

Can we open this issue as a discussion about what can we do to prevent this?

Storing IP's doesn't help AND it's off the table because of (valid) private policy concerns.

Maybe we could have room owners be able to retrieve the IP Address of a _connected_ user and put a (temp) time based ban on their IP address.

Or, the temp ban can be for logging in and/or creating new accounts?

if the app pool restarts, all bans are purged from ram. so this is all temp.

maybe this can be for admins instead of room owners or if a person has a BAN claim instead.

cmaish:
If we can capture the IP addresses of the currently connected users, there could be less of an impact to privacy (doubly so if we hash [and/or salt] them).  I was looking at extending ChatClients for this (it's ephimeral enough that it's not really storing anything more than signalR would have to currently).

If it's definitely a no-go, ban-on-next-message would allow us to avoid collecting IP information from legitimate users.  The hooks could be somewhere around Chat.cs' Send message.  This is a little bit messy though.

Alternatively, a client-side '/ignore' command is a good option - a hellban means that the bot/troll doesn't know if their messages are getting through or not, so they'll likely lose interest sooner.

PureKrome:
interesting idea of letting the bot _in_ so they don't create more accounts but just making them think they are talking and seeing everyone.

but i prefer the bouncer approach personally - nope, not allowed in. F off.

PR #905, opened 2013-06-24
Quick and dirty BanIP command which bans the user and blocks their client IP addresses from future signin

cmaish:
Possibly not appropriate if we can't collect IPs as this captures client ip addresses (currently in plaintext) against ChatClients, and stores abusive IP addresses in plaintext in BannedIPs


cmaish:
Should interim-ly address part of #904

davidfowl:
After discussing this we decided to do a softban feature instead.

Issue #887, opened 2013-06-15
Nudge functionality is broken


cmaish:
As a result of removing jquery-ui, nudge no longer causes the window move effect (in dev).  This is because there's a dependency at:

https://github.com/JabbR/JabbR/blob/master/JabbR/Chat.js#L655

We should reimplement this ourselves, jqui is too heavyweight and doesn't seem to be used elsewhere.

cmaish:
There's a simple pulse implementation given in http://stackoverflow.com/questions/14934336/can-i-get-jquery-pulsate-effect-without-using-jquery-ui that we could adapt and use.

johnbfair:
Great link @cmaish I think we could adapt that and do something similar.  Just want to get @davidfowl to weigh in on it and we can get a fix for this in soon.

davidfowl:
What about https://github.com/jsoverson/jquery.pulse.js

PR #907, opened 2013-06-24
switch pulse target to chat-area because IE (10) doesn't pulse the body on nudges

cmaish:
Other browsers seem to work fine, but IE doesn't pulse anything when the effect is applied to document.body.


cmaish:
Addresses remainder of #887

PR #908, opened 2013-06-24
Add silentban command

cmaish:


JabbR/Commands/BanCommand.cs
```diff
@@ -18,7 +18,7 @@ public override void ExecuteAdminOperation(CommandContext context, CallerContext
 
             ChatUser targetUser = context.Repository.VerifyUser(targetUserName);
 
-            context.Service.BanUser(callingUser, targetUser);
+            context.Service.BanUser(callingUser, targetUser, false);
```
davidfowl: For bool parameters use the optional parameter syntax:

``` csharp
context.Service.BanUser(callingUser, targetUser, silent: false);
```

JabbR/Migrations/201306240509060_EnhanceChatUserBanFlag.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    using System;
+    using System.Data.Entity.Migrations;
+    
+    public partial class EnhanceChatUserBanFlag : DbMigration
+    {
+        public override void Up()
+        {
+            AddColumn("dbo.ChatUsers", "BanStatus", c => c.Int(nullable: false));
+            Sql("UPDATE dbo.ChatUsers SET BanStatus = CASE WHEN IsBanned = '1' THEN '1' ELSE '0' END");
```
davidfowl, line 11: This better work on sql azure :smile:

JabbR/Infrastructure/JabbRFormsAuthenticationProvider.cs
```diff
@@ -120,7 +120,7 @@ public void ResponseSignIn(FormsResponseSignInContext context)
         private static void AddClaim(FormsResponseSignInContext context, ChatUser user)
         {
             // Do nothing if the user is banned
-            if (user.IsBanned)
+            if (user.BanStatus != UserBanStatus.NotBanned)
```
davidfowl: This doesn't look right. We allow logins for silent bans right?

JabbR/Commands/CreateCommand.cs
```diff
@@ -34,11 +34,14 @@ public override void Execute(CommandContext context, CallerContext callerContext
             }
 
             // Create the room, then join it
-            room = context.Service.AddRoom(callingUser, roomName);
+            if (callingUser.BanStatus == UserBanStatus.NotBanned)
+            {
+                room = context.Service.AddRoom(callingUser, roomName);
 
-            context.Service.JoinRoom(callingUser, room, null);
+                context.Service.JoinRoom(callingUser, room, null);
```
davidfowl: Update this call to use the optional parameter syntax as well for null

JabbR/Migrations/201306240509060_EnhanceChatUserBanFlag.cs
nschonni, line 11: `CASE` is available, but `IIF` might be better
http://msdn.microsoft.com/en-us/library/windowsazure/ee336248.aspx#logical

``` sql
UPDATE dbo.ChatUsers SET BanStatus = IIF(IsBanned = '1', '1', '0')
```

cmaish, line 11: IIF is only available on Azure and SQL 2012.  

I know of at least 1 production deployment of Jabbr that uses SQL 2008 R2 (plus my local machine uses it too), so I didn't want to drop that compatibility at this stage (not that I'm particularly married to it working on older infrastructure...)

samandmoore, line 11: Why is this an integer and not a boolean? Is that just for future proofing
if we convert to an enum instead of a binary?

nschonni, line 11: I did read something about that, but I did find this http://msdn.microsoft.com/en-us/library/aa177960(v=SQL.80).aspx as well that seems to say it's been around since at least 2000.

samandmoore, line 11: The case has only 1 when and an else, so it is the same as iff.
:smile:

nschonni, line 11: @samandmoore I usually use `tinyint` for that future proofing, unless you think you'll need more than 256 states

cmaish, line 11: IIf was in SQL Analysis Services 2000 for MDX queries, but I get an error running an SQL Query with IIF in there.

The new BanStatus column is an integer because it maps to an enum in EF - which has NotBanned = 0, Banned = 1, SilentlyBanned = 2.  The old field was a boolean but that wasn't enough.  For the new datatype it's a full integer (rather than a tinyint/byte/etc) because the underlying type of C# enums is a 32-bit int (and while you can override that in the enum definition, there didn't seem to be much point) - also it's what EF's code-based migration builder gave me :P .

nschonni, line 11: Yup, you're right I was thinking of `byte`, but it doesn't make sense since C# using Int32 anyway.
I formally revoke my `IIF` suggestion :smile:

JabbR/Commands/BanCommand.cs
```diff
@@ -18,8 +18,13 @@ public override void ExecuteAdminOperation(CommandContext context, CallerContext
 
             ChatUser targetUser = context.Repository.VerifyUser(targetUserName);
 
-            context.Service.BanUser(callingUser, targetUser);
-            context.NotificationService.BanUser(targetUser);
+            if (targetUser.BanStatus == UserBanStatus.Banned)
+            {
+                throw new InvalidOperationException(string.Format("{0} is already banned.", targetUser.Name));
```
davidfowl: Uppercase S

JabbR/Commands/SilentBanCommand.cs
```diff
@@ -0,0 +1,31 @@
...
+                throw new InvalidOperationException("Who do you want to ban?");
+            }
+
+            string targetUserName = args[0];
+
+            ChatUser targetUser = context.Repository.VerifyUser(targetUserName);
+
+            if (targetUser.BanStatus == UserBanStatus.SilentlyBanned)
+            {
+                throw new InvalidOperationException(string.Format("{0} is already silently banned.", targetUser.Name));
```
davidfowl: percsae S

JabbR/ContentProviders/Core/ContentProviderProcessor.cs
```diff
@@ -44,16 +45,18 @@ public ContentProviderProcessor(IKernel kernel)
                     }
 
                     // Update the message with the content
-
-                    // REVIEW: Does it even make sense to get multiple results?
-                    using (var repository = _kernel.Get<IJabbrRepository>())
+                    if (updateDatabase)
```
davidfowl: Do we still need this since things always get saved?

cmaish:
Requires a breaking migration, so it's a no-go at this stage.  We might revisit this with a change to the user table adding a new column but not killing an existing one at some point in the future...

PR #911, opened 2013-06-25
Allow user to reset their own password

gblmarquez:


JabbR/Models/ChatUser.cs
```diff
@@ -40,6 +40,10 @@ public class ChatUser
         public bool IsAdmin { get; set; }
         public bool IsBanned { get; set; }
 
+        // Password reset id
+        public string RequestPasswordResetId { get; set; }
+        public DateTime? RequestPasswordResetValidThrough { get; set; }
```
davidfowl: DateTimeOffset

JabbR/Nancy/AccountModule.cs
```diff
@@ -342,6 +343,125 @@ public class AccountModule : JabbRModule
...
+                    if (ModelValidationResult.IsValid)
+                    {
+                        ChatUser user = repository.GetUserByName(username);
+
+                        if (user != null)
+                        {
+                            membershipService.RequestResetPassword(user, applicationSettings.RequestResetPasswordValidThroughInHours);
+                            repository.CommitChanges();
+
+                            emailService.SendRequestResetPassword(user, string.Concat(this.Request.Url.SiteBase, "/account/resetpassword/"));
```
davidfowl: Dont' use String.Concat use +

JabbR/Nancy/AccountModule.cs
```diff
@@ -342,6 +343,125 @@ public class AccountModule : JabbRModule
...
+            };
+
+            Get["/resetpassword/{id}"] = _ =>
+            {
+                if (!applicationSettings.AllowUserResetPassword)
+                {
+                    return HttpStatusCode.NotFound;
+                }
+
+                string requestResetPasswordId = _.id;
```
davidfowl: Give _ a proper name please.

JabbR/Services/Email/ContentTypes.cs
```diff
@@ -0,0 +1,8 @@
+namespace JabbR.Services
+{
+    public static class ContentTypes
```
davidfowl: Should be in Intrastructure not Services.

JabbR/Services/Email/FileEmailTemplateContentReader.cs
```diff
@@ -0,0 +1,76 @@
...
+    public class FileEmailTemplateContentReader : IEmailTemplateContentReader
+    {
+        public FileEmailTemplateContentReader()
+            : this("views/emailtemplates", ".cshtml")
+        {
+        }
+
+        public FileEmailTemplateContentReader(string templateDirectory, string fileExtension)
+        {
+            if (string.IsNullOrEmpty(templateDirectory))
```
davidfowl: uppercase S

JabbR/Services/Email/FileEmailTemplateContentReader.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                throw new ArgumentException(string.Format(CultureInfo.CurrentUICulture, "\"{0}\" cannot be blank.", "templateDirectory"));
```
davidfowl: Same here

JabbR/Services/InMemoryRepository.cs
```diff
@@ -1,8 +1,8 @@
-﻿using System;
+﻿using JabbR.Infrastructure;
+using JabbR.Models;
```
davidfowl: You gotta change your VS settings to put System\* namespaces first then change all the files back.

JabbR/Services/InMemoryRepository.cs
```diff
@@ -187,6 +187,13 @@ public ChatUser GetUserByIdentity(string providerName, string userIdentity)
             return null;
         }
 
+        public ChatUser GetUserByRequestResetPasswordId(string requestResetPasswordId)
+        {
+            return _users.FirstOrDefault(u => u.RequestPasswordResetId != null &&
+                                              u.RequestPasswordResetId.Equals(requestResetPasswordId, StringComparison.OrdinalIgnoreCase) &&
+                                              u.RequestPasswordResetValidThrough > DateTime.Now);
```
davidfowl: Always use UTC

JabbR/Services/MembershipService.cs
```diff
@@ -178,6 +178,22 @@ public void ChangeUserPassword(ChatUser user, string oldPassword, string newPass
             EnsureSaltedPassword(user, newPassword);
         }
 
+        public void RequestResetPassword(ChatUser user, int requestValidThroughInHours)
+        {
+            user.RequestPasswordResetId = Guid.NewGuid().ToString().ToSha256(user.Salt);
```
davidfowl: How did you end up using this logic?

JabbR/Services/MembershipService.cs
```diff
@@ -178,6 +178,22 @@ public void ChangeUserPassword(ChatUser user, string oldPassword, string newPass
...
+            user.RequestPasswordResetValidThrough = DateTime.Now.AddHours(requestValidThroughInHours);
```
davidfowl: DateTimeOffset.UtcNow

JabbR/Web.config
```diff
@@ -44,6 +44,13 @@
         <compilation debug="true" targetFramework="4.5"></compilation>
         <httpRuntime requestValidationMode="2.0" targetFramework="4.5" maxRequestLength="524288000" />
     </system.web>
+    <system.net>
```
davidfowl: This obviously can't be checked in and should be off by default.

JabbR/Services/MembershipService.cs
gblmarquez: We need a unique id for each request, that's why I used Guid. I created the Hash with the salt to create a stronger id, but it's not a must. Maybe we could hash a string concat of username and request expiration time.

JabbR/Services/Email/EmailTemplateModelWrapper.cs
```diff
@@ -0,0 +1,44 @@
...
+        private readonly IDictionary<string, object> _propertyMap;
+
+        public EmailTemplateModelWrapper(IDictionary<string, object> propertyMap)
+        {
+            if (propertyMap == null)
+            {
+                throw new System.ArgumentNullException("propertyMap");
+            }
+
+            this._propertyMap = propertyMap;
```
davidfowl: Get rid of the this. references

cmaish:
Most of my feedback is pretty nitpicky and minor - stuff that isn't on its own wrong, just slightly different from how things are done in the existing codebase (for instance String.Formats etc using CultureInfo).  In a lot of cases, you've done it in a more fxcop-ey way - it's a good PR.

That said, there are a couple of things that I'd like to see before this PR gets pulled though:
- The Administrator's JabbR Settings page's "Reset Password Valid Through (Hours)" textbox doesn't populate (even though the value is saved in the database)
- When someone enters a username that does not exist, they get a success message from the reset password screen indicating that an email has been sent.  There's no additional security risk from fuskering valid usernames, so an error is IMO best.
- (Low priority) when things are not successfully configured, ideally there'd either be a validation error on the global settings page or the reset link would not be displayed.  While it's not possible to make sure that everything will work before showing the feautre, ideally messages like "The SMTP host was not specified." would not be displayed to users (if need be, this should be a generic configuration error message that can be localised).
- Not all usernames have associated passwords (twitter social logins).  In some cases, email addresses are associated with the ChatUserIdentity rather than the ChatUser as well.  If we can't send the email message to these users, it's not a huge deal, but "Value cannot be null. Parameter name: addresses" shouldn't be the validation message.
- Expired (outdated, superseded, and consumed) password reset emails show a 404 instead of a useful error message.  If it looks like it could have been a valid link from a real email in the past, we should give some advice to what went wrong.  Users might leave a password expiry email sitting around for a few days for instance and be confused when the links 404.
- GetUserByRequestResetPasswordId searching for a match across the whole database makes me a little uneasy - I'd like to see the reset functionality require a user parameter.  Additionally IIRC, GUIDs aren't particularly random (http://blogs.msdn.com/b/oldnewthing/archive/2008/06/27/8659071.aspx).  I'd prefer to see some RNGCryptoServiceProvider bytes along with something that narrows the search to a single user.  It doesn't need to be hard to reverse (so hashing it isn't necessary).  This one's definitely just my paranoia/background, but I'll probably PR this change later if no one else does it first.

gblmarquez:
Guys I did some refactoring based on feedbacks.

JabbR/Infrastructure/HttpServerUtility.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System;
```
davidfowl: Where did this code come from? Did you copy this from somewhere?

JabbR/Nancy/AccountModule.cs
```diff
@@ -342,6 +344,149 @@ public class AccountModule : JabbRModule
...
+                }
+                catch (Exception ex)
+                {
+                    this.AddValidationError("_FORM", ex.Message);
+                }
+
+                return View["requestresetpassword"];
+            };
+
+            Get["/resetpassword/{id}"] = parameters =>
```
cmaish: you could probably get rid of the token encoding stuff by making this Get["/resetpassword/{user}/{id}"]

gblmarquez: I think it's seems more reliable/robust for the user if we use just the token, by the way all other websites does something similar. But if it's an issue for the project I can refactor it.

cmaish: Put the user into the token or not, it doesn't really matter for all practical purposes so long as the entropy is high enough.  

UUIDs guarantee uniqueness rather than randomness which was where most of my (initial) complaint lay, because you could trigger a couple of resets and potentially figure out what the next reset code would be.

Now that we're using pseudorandom numbers for the token, it's possible (though unlikely, particularly with a longer token-length) that we'll have uniqueness issues (by their nature random numbers can't guarantee uniqueness).  Online services that use just a GUID, or otherwise don't lock the token down to an account when using random but nonunique tokens are doing it wrong IMO.

That said, given the nature of JabbR, all of this is probably getting a little pedantic (which is why whatever approach probably doesn't matter that much).  We're not talking gmail in terms of scale and desirability of getting access, so it's unlikely that a JabbR instance will ever be attacked in this way.

PR #918, opened 2013-07-01
Minor updates to fix graphical glitches / redundancies

cmaish:
remove pointer from userlist headers, update jquery color to make background transitions (eg when the room user count is updated) work, remove unused Lato google web font.


davidfowl:
Need some pics

Issue #928, opened 2013-07-05
AFK doesn't show for people who enter the room after it's set


cmaish:
Have two users in a room and make one /afk - the user list on the other will indicate that the user is idle.

If you connect as a third user, or refresh the non-afk user's page, the afk user will show up as active.

cmaish:
This also affects the message status PR (#929), but a serverside fix to this should resolve both.

PR #929, opened 2013-07-05
Global status indicator as blocky rectangle, user status indicators as blocky rectangles next to messages

cmaish:
Also renames idle throughout UI to inactive (to match underlying user status).

jabbr-status-updates

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -1579,7 +1579,8 @@
         },
         addUser: function (userViewModel, roomName) {
             var room = getRoomElements(roomName),
-                $user = null;
+                $user = null,
+                $userMessages = room.messages.find(getUserClassName(userViewModel.name));
 
             // Remove all users that are being removed
             room.users.find('.removing').remove();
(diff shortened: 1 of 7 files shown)


JabbR/ViewModels/MessageViewModel.cs
```diff
@@ -3,6 +3,8 @@
 
 namespace JabbR.ViewModels
 {
+    using JabbR.Services;
```
davidfowl: usings don't go inside.

JabbR/Chat.ui.js
```diff
@@ -1596,6 +1597,8 @@
             $user.data('owner', userViewModel.owner);
             $user.data('admin', userViewModel.admin);
 
+            $userMessages.find('.user').removeClass('offline active inactive absent present').addClass('active present');
```
davidfowl, line 1600: Looks expensive.

davidfowl:
Does this work when loading history? What about performance concerns? Updating a potentially large number of user messages every time a user's status changes

cmaish:
It works when loading history (though suffers from the issue with https://github.com/JabbR/JabbR/issues/928 if a user has manually afked), and though I don't believe that the performance overhead is going to be that great (it should just cause a single reflow/layout and under the hood is a pretty straightforward selector / .css update operation, perfoemed realistically on less than 1000 DOM elements) and that it'd be dwarfed by the overhead of adding a new message I'll throw together some performance figures when I get the chance.

Might be easiest to wait for #928 to get fixed before pulling this one in.

davidfowl:
Time to see how this does in prod

PR #933, opened 2013-07-07
fix #928 - AFK enhancements

### LINKED ISSUE #928: AFK doesn't show for people who enter the room after it's set

cmaish:
Make afk show when the the room is refreshed.  Add afk handling for message view models.  Make livestamp show "" instead of "s" when we're at 0 minutes.  Show livestamp on manual afks.


JabbR/Chat.ui.js
```diff
@@ -1659,6 +1660,10 @@
             $user.data('active', false);
             $user.addClass('inactive');
             
+            if (!$inactiveSince.html()) {
```
davidfowl, line 1663: What's the deal with this

cmaish, line 1663: It's the pattern used at https://github.com/cmaish/JabbR/blob/6508b25c2c93515660efcbc8087502ae9a02b9da/JabbR/Chat.ui.js#L1625 to only add a livestamp if one doesn't exist.  This change is here to make it so that when you /afk the number of minutes since you went inactive are shown to users who are in the room (to make AFK behave more like every  other inactive).

PR #934, opened 2013-07-07
Notification support for firefox 22+

gscragg:
Redid most of the toast script so there can be (if needed) different implementations. At this stage there is 3:
- the default ( no toast)
- [Html 5 draft](http://www.w3.org/TR/notifications/) support. Although this support is cut down to function exactly like the existing toast, as opposed to the standard.
- Webkit support

### DIFF
--- a/JabbR/Chat.toast.js
+++ b/JabbR/Chat.toast.js
@@ -1,95 +1,151 @@
 ﻿/// <reference path="Scripts/jquery-1.7.js" />
-(function($, window, utility) {
+(function ($, window, utility) {
     "use strict";
 
-    var ToastStatus = { Allowed: 0, NotConfigured: 1, Blocked: 2 },
-        toastTimeOut = 10000,
-        chromeToast = null,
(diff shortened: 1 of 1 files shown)


davidfowl:
Your formatting options put a space before (, we don't do that in any of the other js files, can you change the settings to match the rest of the project?

gscragg:
You mean between function and ()?

davidfowl:
Any ( but yes the examples I see in this change are mostly function(

davidfowl:
hey @gscragg, wanna add support for the new chrome notifications? http://www.theverge.com/2013/7/9/4506814/chrome-notification-center-windows-chrome-os

gscragg:
Yeah, I saw that they had been released.
What kind of functionality are you looking for? 
I was thinking of being able to reply directly to a person who messaged you (if relevant).
Showing users gravatar if it was a direct message (via /msg or your username was mentioned)
Can discuss in #meta if you wish

PR #940, opened 2013-07-11
Improve admin screen layout

cmaish:
Make the labels 100px wider than bootstrap's default, make the labels on the administration screen focus on the field on click.

jabbr-increased-controlwidth

Note that the increased width is present on the user management screens as well.

jabbr-increased-controlwidth_6e72


davidfowl:
Does this affect the login screen in anyway?

cmaish:
It didn't affect the login screen, but the password reset screen/register is structured in a slightly different way to login and so was affected (things were pushed right by 50px).  I've reduced the scope of the selectors in the meantime.

PR #943, opened 2013-07-14
Drag and drop tabs, tab overflow into bootstrap dropdown

cmaish:
Comments or changesets on how I could better deserialise/reserialise user preferences, improve sorting or do things better in general would be much appreciated.  Also looking for discussion in general.  There are likely a couple of rough edges still in there.

jabbr-drag-drop-tabs-and-tab-overflow

New rooms also now open at the end of the room list rather than in alphabetical order (because you can re-sort them yourself).


JabbR/Chat.js
```diff
@@ -269,7 +270,7 @@
     };
 
     // Called when a returning users join chat
-    chat.client.logOn = function (rooms, myRooms) {
+    chat.client.logOn = function (rooms, myRooms, tabOrder) {
```
davidfowl: Change tabOrder to preferences. We'll generalize and pass a settings object instead of just the tab order.

JabbR/Chat.js
```diff
@@ -289,6 +290,10 @@
                 ui.setRoomClosed(room.Name);
             }
         });
+
+        chat.state.tabOrder = tabOrder;
```
davidfowl: Does it need to be on state?

JabbR/Models/ChatUser.cs
```diff
@@ -71,5 +78,36 @@ public bool HasUserNameAndPasswordCredentials()
         {
             return !String.IsNullOrEmpty(HashedPassword) && !String.IsNullOrEmpty(Name);
         }
+
+        [NotMapped]
+        public ChatUserPreferences Preferences
+        {
+            get
+            {
+                if (this.RawPreferences == null)
```
davidfowl: No this.

JabbR/Hubs/Chat.cs
```diff
@@ -551,7 +569,7 @@ private void LogOn(ChatUser user, string clientId, bool reconnecting)
                 }
 
                 // Initialize the chat with the rooms the user is in
-                Clients.Caller.logOn(rooms, privateRooms);
+                Clients.Caller.logOn(rooms, privateRooms, user.Preferences.TabOrder);
```
davidfowl: See other comment, pass the preferences object.

JabbR/Chat.js
cmaish: no, and I should probably move other things like unreadNotifications off at the same time as well... or we could change the cookie to be somewhere else and contain only the active room and anything else that we want to be able to grab via a context at the server.

Issue #948, opened 2013-07-17
htmley characters on notification screen are incorrectly escaped


cmaish:
Message text containing < renders as &amp;lt; on the notification screen for instance.

Seems to be related to #952.  It's likely that a fix would address both issues.

cmaish:
Fixed and in prod

PR #953, opened 2013-07-22
Fix #952 and #948 by encoding metadata for notifications

### LINKED ISSUE #952: Notifications show HTML
image

cmaish:
Fix #952 and #948 by encoding metadata for notifications in different ways


cmaish:
Should also fix #948 now.

Issue #950, opened 2013-07-21
times on messages are sometimes wrong


cmaish:
Some messages are using local time by the looks of things - which can result in the message times being out of order and off (potentially be a decent amount):

![jabbr-whatisthis](https://f.cloud.github.com/assets/1271535/832215/c457a2d4-f23b-11e2-9346-954dab1d7320.png)

cmaish:
Now we update the time when we roundtrip messages from the server.  I get the impression that there are still some edge cases of this around (I don't think that /msg notifications will get the updated time but don't have a setup to test right now).

garysharp:
Confirmed that this is still an issue for 'notification' messages:
https://github.com/JabbR/JabbR/blob/ff88743eb61eb9e09544c4b73a81f2f76bbfb4ff/JabbR/Chat.ui.js#L1858
Includes private messages, room entry/exit, etc.

There are 5 other places that use the local date in Chat.ui.js.

I'm not sure how this should be corrected. A few options:
1. Detect the time difference and notify the user. Then leave it to the user to correct (update the time on their client).
2. Maintain a 'time offset', and then apply the correction to all client-generated dates.
Note: Time difference could be detected with a 'getServerTime' hub method - but assumes relatively low-latency connection.

There is also another code-path with messages from the server that don't take the opportunity to update the time, but I can't see this being used at the moment (unless messages got sent twice, etc): ui.replaceMessage
https://github.com/JabbR/JabbR/blob/ff88743eb61eb9e09544c4b73a81f2f76bbfb4ff/JabbR/Chat.js#L414
https://github.com/JabbR/JabbR/blob/ff88743eb61eb9e09544c4b73a81f2f76bbfb4ff/JabbR/Chat.ui.js#L1820
(Its not even used in 'chat.client.replaceMessage' - that uses 'ui.overwriteMessage')

PR #958, opened 2013-07-28
Update client-sent messages with server time

garysharp:
Should resolve issue #950

Caused by the client and server's time being out-of-sync, this PR updates the client's view of the message with the time recorded by the server.

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -1817,6 +1817,7 @@
             processMessage(message);
 
             $message.find('.middle').html(message.message);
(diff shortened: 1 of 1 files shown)


garysharp:
Just a thought... Should JabbR warn the user to correct their time if it is out of sync by XX amount? This problem (caused by the end-user presumably, assuming that JabbR's host is correctly synced) might still cause grief if the user is trying to compare a JabbR conversation with some other external resource (eg. the system clock). I suppose it would be more of a 'heads-up' for the user (a community service) rather than providing any huge benefit.

Issue #309, opened 2012-01-23
Display the AFK Note inline, if one has been set.


PureKrome:
When a person goes afk, they currently have the option to set some afk text.

eg.

/afk going to lunch
/afk meetings - no idea when i'll be back.

i don't think people know that 
1) u can actually set an afk note
2) if someone DOES set the afk message/note, to actually check on that.

So i was thinking that when an afk message/note has been set, the UI displays the following:-

XXX has gone AFK, with the message: YYYY ZZZ ...

or something like that .. at least the message/note is displayed there. Of course, the afk note remains in the ChatUser listing paperclip, etc (all that stuff is unchanged).

I'm happy to do this, btw. If u like this idea, just confirm the text required.
otherwise, close away :)

davidfowl:
@PureKrome send a PR for this one.

PR #962, opened 2013-07-28
Improve AFK/Note

cmaish:
Make AFK Note / Note display what they've been set to (if set) to both you and everyone else (implement #309).  Tidy up related code.


cmaish:
Rebased on current dev.

PR #964, opened 2013-07-30
Fix tabheaders

### LINKED ISSUE #960: Header tabs sometimes remain highlighted after being viewed
jabbr-header-highlight

At the moment, we add the highlight using javascript animations, which adds the background colour on the element itself.  We should just add/remove a class with the unread messages (we can use css animations if they're available)

cmaish:
Remove animation on adding unread class to room header tabs, or fix #960.  I played around with CSS animations, but they don't seem powerful enough to handle what we want (fade-in but with interruption on hover) without adding extra dom elements.

### DIFF
--- a/JabbR/Chat.ui.room.js
+++ b/JabbR/Chat.ui.room.js
@@ -3,36 +3,6 @@
     
     var trimRoomHistoryMaxMessages = 200;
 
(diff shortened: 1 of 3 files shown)


davidfowl:
No more animation?

PR #982, opened 2013-09-11
I don't always get new commands, but when I do, it's a MEME command.

PureKrome:
Hi David. First pass at adding a simple meme command.
### MEME COMMAND

usage: `/meme <meme-type> <top-line> <bottom-line>`
eg: `/meme aa Why-is-JabbR-offline? Aliens`

Included are some unit tests and a simple method which ends up calling the `Send(..)` method.

SIDE QUIRK: I noticed the test classes in `JabbR.Tests` where missing the 's' so I added that.


DavidChristiansen:
Why so many checkins!? Did you fetch/pull/stretch/relax/pontificate prior to pushing?

PureKrome:
:cry: I donno. i just forked Jabbr. then switched to dev. did code. when to push and it said i was behind. pulled. tested again. then pushed :cry: :cry: :cry:

I just thought that was the state of upstream/dev ?

damianh:
This might be the most important feature added to Jabbr since it was moved to azure for scale out and stability purposes.. oh wait

DavidChristiansen:
I propose the command be

```
/meme <graphic> <First Line, single word or quoted> <Second Line, single word or quoted>
```

example

```
/meme aa "Why is JabbR down again?" Aliens
```

DavidChristiansen:
Also, I would try fix this PR to be JUST your checkins, otherwise you have little chance of it being accepted.

Buildstarted:
The dashes instead of spaces is very wrong. Users should not be required to conform to upboat.me just to add a meme, it should be much more natural. a single separator or quotes should be used to delineate between lines.

and-you-should-feel-bad

davidfowl:
Wow, epic fail PR. Redo.

PureKrome:
Ok guys. For the "top line" "bottom line", gimme some code that JOINS the space delimited string (args) then splits up into those parts, looking for a start and end quote.

PureKrome:
Bah. it's 1:40am. :crescent_moon:  I tried. failed. Someone else do it instead.

I suck

damianh:
@PureKrome Lrn2rebase (I guess)

PR #989, opened 2013-09-17
Fix text topic title

tieungao79:
untitled

Fix that proplem.


cmaish:
Seems that https://github.com/JabbR/JabbR/pull/999 is the new version of this PR (which does't have db instance changes / commented out stuff in CSS).  Closing for now.

PR #993, opened 2013-09-18
Make initial load seem more smooth

cmaish:
At current, when a user accesses JabbR, the page and scripts are loaded (5k), then the key combinations are loaded (<1k),  then the entire list of rooms (~105kb) is loaded, then the current room's contents are loaded (usually around 25kb), then each other room is loaded one at a time.

This change rearranges the load of things to be: your current tab then the lobby, other rooms (as a batch) and commands.  It makes things feel quite a bit faster, particularly down slow connections, even though we're not doing any less work.


JabbR/Hubs/Chat.cs
```diff
@@ -377,6 +377,20 @@ public Task<List<LobbyRoomViewModel>> GetRooms()
                                    .Select(m => new MessageViewModel(m));
         }
 
+        public async Task<RoomViewModel[]> GetRoomsInfo(string[] roomNames)
+        {
+            // can't async whenall because we'd be hitting a single EF context with multiple concurrent queries.
+            var roomInfo = new List<RoomViewModel>();
```
davidfowl: This sucks but is interesting. Can instead call the client back as results are ready instead of waiting on this call to return?

``` C#
foreach (var roomName in roomNames)
{
    var roomInfoVal = await GetRoomInfo(roomName);
    Clients.Caller.roomLoaded(roomInfoValue);
}
```

PR #996, opened 2013-09-18
Fix #994 - the display of second page of rooms on lobby screen was broken

### LINKED ISSUE #994: Lobby "Load More" seems to skip some rooms.
On my local dev install, I created 1000 rooms "Room1" to "Room1000".  Rooms appear in alphabetical order, with 188 being the last room that's displayed in the list when I load the lobby.

When I click "Load more rooms" the next room is 279 - I cannot seem to find rooms 189 - 278.

Note that this is just the first "Load More" click.  Subsequent clicks continue on properly.

cmaish:
When you clicked Load More on the lobby screen, it would skip the second page of results (results 101-200) and skip straight to 201-300.  Subsequent clicks would however work.

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -1473,7 +1473,7 @@
                 }
 
                 var listOfRooms = $('<ul/>');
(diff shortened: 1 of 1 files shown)


davidfowl:
This looks fine.

PR #999, opened 2013-09-18
Fix text topic title when resize 

tieungao79:
Before change
untitled
After change
untitled1


JabbR/Content/themes/default/Chat.css
```diff
@@ -228,6 +228,9 @@ form#send-message {
 .roomTopic {
     position: absolute;
     margin-left: 5px;
+    width: calc(100% - 123px);
```
cmaish: Is Safari a browser we need to support?  IIRC we need -webkit-calc( for it to work there.  "left:0; right:123px;" would be equivalent in this case.

For overflow, we can add text-overflow:ellipsis; to get the "..." when a word is truncated, and white-space:nowrap; to prevent words from going to the (hidden) next line without the ellipsis.

JabbR/Content/themes/default/Chat.css
```diff
@@ -806,8 +806,13 @@ li.user.typing .gravatar {
     position: absolute;
     right: 0;
     width: 70px;
+	height: 30px;
```
davidfowl, line 809: Fix the formatting. Make sure you're not using tabs.

davidfowl:
@alfhenrik and @cmaish what say you?

hnrkndrssn:
I'm confused, the pull request is for fixing the room topic, but the changes in the pull request for fixing the room topic has been overwritten by https://github.com/tieungao79/JabbR/commit/567e97af97af27a0b8290405e8474e4cf7a5b8fe...which is for fixing display issue of the upload/send buttons in FF...or am I seeing things?

cmaish:
jabbr-longtopic

The changes in this PR don't actually handle overflow on the topic at the moment.  They affect the height of the send button and seem to ensure that there's a specific line height when icons are added.

davidfowl:
So this change isn't good then?

hnrkndrssn:
Well, it might be good for fixing the Upload/Send buttons in FF :smiley: , but for fixing the Room Topic when resizing the window, no.

cmaish:
This is deprecated as it's fixed in #1045

Issue #991, opened 2013-09-18
Room autocomplete (#RoomNam) is flaky


cmaish:
Autocomplete for rooms currently looks at the list items that are on the lobby page.

It should instead look at chat.ui's roomCache (or a case-correct equivalent).

To test, enter #met and press tab (it should find meta).  Now, try the same with a private room of yours; or a room that isn't currently displayed (either as a result of filtering via the search box or as a result of being off the page in terms of loading) - it is not found.

Issue #992, opened 2013-09-18
Autocomplete values are not sorted (other than emoji)


cmaish:
These should probably be alphabetically sorted.

When you type @ then press tab, it gives users in display order, when you type # then press tab, it gives rooms in the order in the lobby, when you type / and press tab it gives commands in a random order.

This also has the related side effect of the help commands not being in any particularly useful order on the help screen.

cmaish:
In dev, should find its way to prod soon.

PR #1000, opened 2013-09-19
Make all autocomplete options sorted / fix room autocomplete

cmaish:
At the moment, autocomplete stuff isn't always sorted.  Typing #RoomName where there are multiple matches doesn't always put them in alphabetical order, or show all rooms.  Users aren't in alphabetical order.  Commands aren't in alphabetical order (either in autocomplete or the help).

This fixes that (#992)


cmaish:
Also fixes #991

PR #1013, opened 2013-09-30
Escape meme

### LINKED ISSUE #1012: Meme integration doesn't handle valid meme's with ?
The following meme path (used via the /meme command)

https://upboat.me/angry-chef-gordon-ramsay/cold-hands?/type-faster!.jpg

generates a valid meme, but does not display in jabbr.  Remove the ? and it does.

cmaish:
Fixes #1012 - encode a bunch of characters (like question marks, slashes, hash characters) to make meme generation better.


hnrkndrssn:
Looks good to me

davidfowl:
+1

PR #1017, opened 2013-10-04
Display splash screen while loading rooms after joining JabbR

hnrkndrssn:
image


JabbR/Chat.js
```diff
@@ -1239,7 +1241,7 @@
 
         // Initialize the ui, passing the user preferences
         ui.initialize(state.preferences);
-
+        //ui.showSplashScreen();
```
davidfowl: Remove dead code

PR #1020, opened 2013-10-07
defer writing messages and handing commands from the hub to the ui tier until the room in question is ready.

cmaish:
Should\* work.  Does stuff like ensures that we don't dup messages that get received between when the client calls logOn and when they actually getRoomInfo (and makes sure that the non-dup messages actually appear correctly in history).

*Should work at least as well as what's in there at the moment.  Though it's less performant because we closure all over the place

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -1958,27 +1958,24 @@
                 $middle.append(content);
             }
         },
-        getAllOpenRooms: function() {
-            var rooms = getAllRoomElements();
+        getOpenRoomNames: function() {
+            var openRooms = $.grep(getAllRoomElements(), function(elem) {
+                return (elem.getName() !== undefined && elem.isClosed() === false);
+            });
 
-            for (var r in rooms) {
-                if (rooms[r].getName() !== undefined && rooms[r].isClosed() === false) {
-                    
-                }
-            }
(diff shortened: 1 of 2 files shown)


JabbR/Chat.ui.js
```diff
@@ -722,6 +735,14 @@
         // re-filter lists
         $lobbyRoomFilterForm.submit();
     }
+    
+    function getRoomReadyDeferred(roomName) {
+        if (typeof roomReadyDeferreds[roomName] === 'undefined' || roomReadyDeferreds[roomName] === null) {
+            roomReadyDeferreds[roomName] = $.Deferred();
```
davidfowl, line 741: Should this be:

```
roomReadyDeferreds[roomName] = roomReadyDeferreds[roomName] || $.Deferred()
```

JabbR/Chat.ui.js
```diff
@@ -1934,11 +1958,25 @@
                 $middle.append(content);
             }
         },
+        getAllOpenRooms: function() {
+            var rooms = getAllRoomElements();
+
+            for (var r in rooms) {
+                if (rooms[r].getName() !== undefined && rooms[r].isClosed() === false) {
```
davidfowl: ?

cmaish:
Kinda fugly.  We can do better when we refactor some js.

Issue #1022, opened 2013-10-08
Room nudge displays incorrect message.


cmaish:
If user1 nudges room room1, and user2 is in room1 and room2 with room2 active, they currently receive a message of "_user1 nudged the room." in room2.  Message should probably be "_user1 nudged room1." and should appear in _all_ rooms for user2...

cmaish:
Fixed in dev.

PR #1025, opened 2013-10-10
Improve room nudge message and behavior.

cmaish:
Send more descriptive messages:

*admin nudged room Canon.
*admin nudged CJM.

And send them to all rooms as private messages.


cmaish:
Oh, fixes #1022

Issue #1007, opened 2013-09-27
When a room is closed, tab focus always goes to lobby.


cmaish:
This should only occur if the current tab is the one being closed.

Replication: Open Lobby, RoomA, RoomB.  Select RoomB.  Close RoomA.  The user gets redirected to the lobby instead of staying in the currently selected room (RoomB).

cmaish:
The same is true when the user is kicked from a room.  Even if it's not the active room, they're moved to the lobby.

PR #1026, opened 2013-10-10
Fix closing to lobby

cmaish:
When you close a tab we always go back to the lobby.  This is silly, and should only happen when you close the currently active tab.  This is issue #1007


davidfowl:
+1

PR #1030, opened 2013-10-16
Update infrastructure

cmaish:
We gotta live on the edge, yo.  Also, myget packages for nightlies disappeared again.


davidfowl:
Looks good

PR #1032, opened 2013-10-21
Fix being able to get to negative unread notifications count.  Fixes #1031

### LINKED ISSUE #1031: Able to get to negative unread message count.
Click the read button on notifications many times and the count continues to decrease.

clipboard_6602 1

cmaish:
When you click on the "I've read this" button, it decrements the unread notifications count by 1.  You can currently click on the button multiple times while the notification is hiding.  This change marks the notification read thing as disabled preventing the second click from triggering the decrement event.

### DIFF
--- a/JabbR/notifications.js
+++ b/JabbR/notifications.js
@@ -16,6 +16,7 @@
         if ($this.hasClass('disabled')) {
             return false;
         }
+        $this.addClass('disabled');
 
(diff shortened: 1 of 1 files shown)


davidfowl:
How?

cmaish:
Added some detail to the PR description - I add the disabled class to the button as soon as it's clicked so that multiple clicks don't cause multiple decrements to the total unread message count.

PR #1033, opened 2013-10-21
Tidy javascript for UI refactoring

cmaish:
Each changeset has its own minor cleaning/refactoring.  

The overall goal is to start to simplify the js code and refactor rooms etc into something that can be extended and used in more general ways.

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -2039,13 +2039,21 @@
         addListToActiveRoom: function(header, messages) {
             this.addList(header, messages, getActiveRoomName());
         },
-        addBroadcast: function(message, roomName) {
-            this.addMessage(message, 'broadcast', roomName);
+        addBroadcast: function (message) {
+            //TODO add some behavior if the lobby is active and/or no rooms are joined
+            var rooms = getAllRoomElements();
+            for (var r in rooms) {
+                // broadcasts are relevent in *all* rooms, even closed ones
+                if (rooms[r].getName() !== undefined) {
+                    this.addMessage(message, 'broadcast', rooms[r].getName());
+                }
+            }
         },
         addAction: function(message, roomName) {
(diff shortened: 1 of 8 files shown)


JabbR/Chat.ui.js
```diff
@@ -1357,13 +1298,13 @@
 
             loadRoomPreferences(roomName);
 
-            if (room.isActive()) {
+            if (room.isCurrent()) {
```
hnrkndrssn, line 1301: Might just be me but, here we use .isCurrent but when retrieving the room elements for the current room we use an 'Active' notion (i.e. getActiveRoomElements). 

I think maybe we should agree on using either a 'Current' or an 'Active' notion? Discuss. :smiley:

cmaish, line 1301: In the future, we want to potentially be able to display multiple chat areas at a time (for splitscreen, for user-user chat, etc).  The convention that I was using is:

"Current" - Means that a room is visible
"Active" - Means that the room has focus

That is, in the future, multiple rooms could be 'current' and we could update logic to only ensure that messages are added to the dom or otherwise displayed on current rooms.  A maximum of 1 room can be active at a time however (when we send messages to the server (even commands, which might come from some alternative UI), they're associated with a room)

Happy for any convention, but that's what my split of Active/Current was working towards.

hnrkndrssn, line 1301: In that case, I agree!

hnrkndrssn:
Not sure what state this is currently in, in regards to being up-to-date with the latest dev branch, but this all looks good to me, and is a good start at getting the js code more manageable :grinning:

cmaish:
It's probably a bit out of date, though I don't think that too many major changes have been made in the js in the past couple of months.  Hopefully it's just a case of a quick rebase...

davidfowl:
@cmaish Can you prepare this to merge

PR #1035, opened 2013-10-25
Add a maximum length administration setting that gets applied to chat messages.

cmaish:
Reject all messages longer than this value (as set in admin settings).  Adds maxlength attribute on clientside but doesn't polyfill for IE9 and earlier (these will be rejected at the server).


JabbR/Services/ApplicationSettings.cs
```diff
@@ -79,6 +81,7 @@ public static ApplicationSettings GetDefaultSettings()
                 EncryptionKey = CryptoHelper.ToHex(GenerateRandomBytes()),
                 VerificationKey = CryptoHelper.ToHex(GenerateRandomBytes()),
                 MaxFileUploadBytes = 5242880,
+                MaxMessageLength = 0,
```
davidfowl, line 84: Pick a better default.

cmaish, line 84: I'm guessing that -1 isn't much better?  

Problem is that when a key's not present but we have settings in the DB it will come in as 0 (the defaults are only used when we can't deserialize for whatever reason).  I think I'll need to write a migration to support this.

PR #1041, opened 2013-10-30
Correct server/client time offset

garysharp:
Server notifies client of time at hub connection and an offset is calculated. Any client created dates are corrected by the calculated offset. Possible solution to #950


davidfowl:
Can you rebase this? It's been like a year I know 😢

garysharp:
Replaced with #1240

PR #1042, opened 2013-10-30
Client warning on significant time difference

garysharp:
Notifies the client when a significant time difference exists between the server and client (>5 seconds). Possible solution to #950

jabbrclienttimeerrormessage


garysharp:
Replaced with #1240

PR #1043, opened 2013-10-30
Fixed name matching regex for mentions

Buildstarted:
- Original name matching regex only replaced the first
  . character instead of all of them


hnrkndrssn:
Looks good to me.

davidfowl:
:shipit:

PR #1045, opened 2013-11-03
Make smaller resolution devices behave better

cmaish:
JabbR is currently not functional on most non-PC devices.  These changes try to improve the experience on devices with low resolutions.
- rejigged the way that the chat screen is structured to better support easier resolution changes (and multi pane), 
- fixed the overflow on the topic when it's too long, 
- fixed the tabs list to support the case where no non-dropdown pages can be shown, 
- added low-res media queries for the main chat window (both vertical and horizontal) which squeeze a bit more stuff into the window (things include making the time display floated right so there's more room for the messages, taking away unnessecary padding, and replacing the JabbR text with the jabbr 32px logo)
- simplified login and account pages (added a span6 to everything, removed now-unnecessary per-page styles) and added low-res media queries,
- made minor changes to the administration pages (things like text overflow in validation key on admin page, adding mobile initial scale meta for other pages)

capture 28
capture 27
capture 25
capture 24
capture 23
capture 22
capture 21
capture 20
capture 19
capture 18
capture 17
capture 16
capture 15
capture 14
capture 13
capture 12
capture 10

### DIFF
--- a/JabbR/Chat.ui.fileUpload.js
+++ b/JabbR/Chat.ui.fileUpload.js
@@ -198,20 +198,7 @@
 
     $(document).on('dragenter dragover', '.messages.current', function () {
         //show drag target
-        //get css position
-        //width,height
-        var position = $(this).offset();
-        var size = { width: $(this).width(), height: $(this).height() };
-        $('#drop-file-target').css({
-            top: position.top + 1,
-            left: position.left + 1,
-            width: size.width,
-            height: size.height,
-            position: 'absolute',
-            background: '#000',
-            opacity: '.25'
-        });
-
+        // TODO: make one of these per room
         $('#drop-file-target').fadeIn(500);
     });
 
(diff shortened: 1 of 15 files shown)


davidfowl:
I approve. @alfhenrik take a look

hnrkndrssn:
Nitpicking here, and we might not care (as that is running 480x800 on desktop), but...
image

Could be fixed by making the right margin -140px instead of -180px.

Other than that looks good! 
:shipit:

cmaish:
Fix'd.  There are probably still a couple of areas where small-device support isn't perfect (notification screen comes to mind as not being amazing when small) but hopefully this is enough to start getting people back using jabbr on mobile devices.

capture 31
capture 30
capture 29

hnrkndrssn:
:shipit:

davidfowl:
Merge this bad boy

PR #1048, opened 2013-11-07
Add SQL scaleout for signalr

cmaish:
Only supporting the azure service bus backplane for scaleout is lame (and means that we can't test locally).


davidfowl:
Get rid of the binding redirect and use the same version as the service bus provider

cmaish:
k, PR should suck less now.

davidfowl:
This is good now. :shipit:

PR #1049, opened 2013-11-12
Client-side Dialogs

garysharp:
Provides server-side method of invoking client-side dialogs, and (initially) a `LockCommand` command which implements the dialog:

jabbr-lockcommanddialog


cmaish:
To me, this sort of change shouldn't have a major effect on the server - it's a UI-level concern.  There's definitely some mixing of levels in the existing jabbr codebase, but my personal approach would have been to send down the confirmation (along with any other data required in the future, such as command builder dialog info) to the client in the bundle of data sent to populate the command autocomplete (so it would have gone into the command attribute).

This way, settings such as confirmations and prompts can be managed at the client... what do you reckon?

davidfowl:
I was thinking the same thing. It's purely a client feature.

hnrkndrssn:
I agree this should be a client side feature, shouldn't need to hit the server to determine if the user have confirmed they want to lock the room or not.

garysharp:
I've tried to take your feedback on-board and rebuilt the PR, moving the logic to the client.

An extra property on the `CommandAttribute` being set indicates confirmation is required. Localized confirmation messages are sent to the client along with the autocomplete data (as per @cmaish recommendation). Before hitting the server, command messages are tested for confirmation and a new argument to the 'sendMessage' event verifies confirmation has been granted (if required).

The UI has changed slightly: _buttons cannot be changed, and the dialog heading has a fixed format._
image

JabbR/Chat.ui.js
```diff
@@ -2064,6 +2065,13 @@
         getCommands: function () {
             return ui.commands || [];
         },
+        getCommand: function (name) {
+            var commands = ui.getCommands().filter(function (c) { return c.Name === name; });
```
davidfowl: Maybe we should store commands by name.

JabbR/Chat.js
```diff
@@ -20,8 +20,7 @@
         messageSendingDelay = 1500,
         pendingMessages = {},
         privateRooms = null,
-        roomsToLoad = 0,
-        commandNameLookup = {};
```
garysharp, line 24: Command lookups migrated to `chat.ui.js`, no longer stored in `chat.js`

JabbR/Chat.ui.js
```diff
@@ -2161,6 +2231,12 @@
         getUserName: function () {
             return ui.name;
         },
+        getUserHash: function () {
```
garysharp, line 2232: Necessary to store (and update) the user's hash to display messages (see line 619)

davidfowl:
I forget where we left this

cmaish:
The dialogs are currently functional and are added via an extra item that references a confirmation message key.

There are three gaps that we may want to look at addressing once we pull in this PR:
- The lock command confirmation always says "this room", even when you run it with /lock otherroomname.  We should look at sanitizing the input string then using the Chat.utility.format function.
- The confirmation shows regardless of whether you can actually perform the action (this is by design - the confirmation is completely client-side, nothing gets sent to the server if the user doesn't confirm).
- There's an amount of markup in the language resource.  This probably isn't an issue because we don't have (m)any localisations.

Otherwise, looks very solid.

garysharp:
With regards to:

> The lock command confirmation always says "this room", even when you run it with `/lock [OtherRoomName]`.

Do you want me to investigate implementing this _before_ the pull?

cmaish:
I'd vote for pulling it in then beefing it up/rolling it out to the other appropriate commands as another PR after it's in the codebase...

hnrkndrssn:
:+1:, let's get it in, and then we can enhance it after that.

hnrkndrssn:
clap-transformers-bumblebee

PR #1052, opened 2014-01-07
Nudge Notifications

mrb0nj:
When nudging a User play an audio and/or display toast notification if enabled in any room. This should grab the users attention when out of browser.

I'm fairly new to JabbR so not sure if this is the best way? feedback welcomed :)


cmaish:
Nice PR - Clean and works well.  Based on how the code is structured at the moment, it's the right way to do things too, IMO.

davidfowl:
Looks good.

hnrkndrssn:
thank-you-spice

mrb0nj:
Many thanks, glad I was on the right track :)

PR #1056, opened 2014-02-01
Refactored LINQ to shorter, faster statement in PersistedRepository.IsUs...

BrianRosamilia:
I don't see any reason why this statement can't be refactored as below.

There are no integration tests to support my claim, but there is no explicitly assigning null to room name anywhere in the project--for this application's purpose, these should work the same from what I can see (and the manual testing I did).

An absolutely equivalent statement should be .Any(r => r.Key == room.Key && r.Name != null) which is still more concise and only returns a bit from the db--rather than the full name of the room (which is totally unnecessary here, even if my PR isn't the optimal refactor :scream:).  

Curious as to whether or not this is an EF optimization trick, but there are usually comments in the codebase for such quirks.

I wouldn't obsess over this if it weren't a very frequent call.

### DIFF
--- a/JabbR/Services/PersistedRepository.cs
+++ b/JabbR/Services/PersistedRepository.cs
@@ -295,9 +295,7 @@ namespace JabbR.Services
             return _db.Entry(user)
                       .Collection(r => r.Rooms)
                       .Query()
-                      .Where(r => r.Key == room.Key)
-                      .Select(r => r.Name)
-                      .FirstOrDefault() != null;
+                      .Any(r => r.Key == room.Key);
         }
 
         public void Reload(object entity)


cmaish:
It may have been an EF optimisation trick.  The queries that are generated for me are quite different between the version in dev and this version.

For the PR version, calling Any():

```
exec sp_executesql N'SELECT 
    CASE WHEN ( EXISTS (SELECT 
        1 AS [C1]
        FROM [dbo].[ChatUserChatRooms] AS [Extent1]
        WHERE ([Extent1].[ChatUser_Key] = @EntityKeyValue1) AND ([Extent1].[ChatRoom_Key] = @p__linq__0) AND (@p__linq__0 IS NOT NULL)
    )) THEN cast(1 as bit) WHEN ( NOT EXISTS (SELECT 
        1 AS [C1]
        FROM [dbo].[ChatUserChatRooms] AS [Extent2]
        WHERE ([Extent2].[ChatUser_Key] = @EntityKeyValue1) AND ([Extent2].[ChatRoom_Key] = @p__linq__0) AND (@p__linq__0 IS NOT NULL)
    )) THEN cast(0 as bit) END AS [C1]
    FROM  ( SELECT 1 AS X ) AS [SingleRowTable1]',N'@EntityKeyValue1 int,@p__linq__0 int',@EntityKeyValue1=1,@p__linq__0=3
```

For the dev version, calling FirstOrDefault() != null:

```
exec sp_executesql N'SELECT 
    [Limit1].[Name] AS [Name]
    FROM ( SELECT TOP (1) 
        [Extent2].[Name] AS [Name]
        FROM  [dbo].[ChatUserChatRooms] AS [Extent1]
        INNER JOIN [dbo].[ChatRooms] AS [Extent2] ON [Extent1].[ChatRoom_Key] = [Extent2].[Key]
        WHERE ([Extent1].[ChatUser_Key] = @EntityKeyValue1) AND ([Extent2].[Key] = @p__linq__0) AND (@p__linq__0 IS NOT NULL)
    )  AS [Limit1]',N'@EntityKeyValue1 int,@p__linq__0 int',@EntityKeyValue1=3,@p__linq__0=3
```

The old query does end up with a simpler (looking, at least) execution plan (see below), with a marginally lower estimated subtree cost.  As a result, it'll probably execute a tiny bit faster on the database.  Sure, the result is a string, but it's a short one, and the size of the initial query and all the other overheads dwarfs it.  All in all, while the change does make the code more concise, I wouldn't expect it to have any measurable impact on performance one way or the other.

executionplan

Additionally, the IsUserInRoom method shouldn't be getting called very frequently - unless you've modified your copy of jabbr, IsUserInRoom should actually be relatively infrequently called.  It's only caller is https://github.com/JabbR/JabbR/blob/dev/JabbR/Models/RepositoryExtensions.cs#L70 .  We have some (aggressive) caching sitting in front of the method.  

TL;DR I don't have anything against pulling in the change personally, though IMO it may be best until we mess around with the repository before pulling it in, just to reduce churn.

BrianRosamilia:
I noticed the query plan getting worse in my solution too, but wanted someone to check my work and draw their own conclusions.  I think you nailed it, but a comment in the codebase would have been good on that method since it's not idiomatic (looking) LINQ.  It would be interesting to check this against future versions of EF.

I may not have grasped the extent of the caching because I was testing in short runs.  Woops; thanks for clarifying that.

I'll agree it's a pretty big micro-optimization of a PR but it was interesting enough for me to bring up.

PR #1058, opened 2014-02-06
Changed Regex to allow pulls to show as well as issues

khellang:
Since pulls are essentially issues, this should show pull requests just like issues.


davidfowl:
Send this to dev

PR #1060, opened 2014-02-06
Changed Regex to allow pulls to show as well as issues

khellang:
This will allow pull requests to be shown like issues.

### DIFF
--- a/JabbR/ContentProviders/GitHubIssuesContentProvider.cs
+++ b/JabbR/ContentProviders/GitHubIssuesContentProvider.cs
@@ -10,7 +10,7 @@ namespace JabbR.ContentProviders
 {
     public class GitHubIssuesContentProvider : CollapsibleContentProvider
     {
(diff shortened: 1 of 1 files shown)


davidfowl:
Did you test this?

khellang:
Yes. Both PR and issue data can be fetched through the issues API, it returns the same data for both:

Issue data for this PR: https://api.github.com/repos/JabbR/JabbR/issues/1060
Issue data for an issue: https://api.github.com/repos/JabbR/JabbR/issues/1055

If you try to paste a PR URL, it won't show the content today (because the Regex doesn't match), but if you replace the `pull` part with `issues` and paste it in JabbR and it works fine :smile:

Now if you use the PR API, you will get a lot more data, e.g. source- and target branch etc., but for now, I think it's OK to show PRs the same as issues.

Pull data for this PR: https://api.github.com/repos/JabbR/JabbR/pulls/1060

PR #1077, opened 2014-03-23
Fix in room cache

cmaish:
Added some backplane communication for server-server communication.  Currently in a rough (but working) state.

Currently adds a singleton ChatServiceProxy class which both subscribes to and publishes events to the backplane communication (which is a singleton wrapper around the signalr message bus).  ChatService calls these methods when necessary.

### DIFF
--- a/JabbR/Infrastructure/BackplaneChannel.cs
+++ b/JabbR/Infrastructure/BackplaneChannel.cs
@@ -1,0 +1,200 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+
+using Microsoft.AspNet.SignalR.Infrastructure;
+using Microsoft.AspNet.SignalR.Json;
+using Microsoft.AspNet.SignalR.Messaging;
+
+using Newtonsoft.Json;
+
+namespace JabbR.Infrastructure
+{
+    public class BackplaneChannel : ISubscriber, IBackplaneChannel
+    {
+        private const string BackplaneSignal = "__BACKPLANE__";
+
+        private readonly List<string> _signals = new List<string>{ BackplaneSignal };
+
+        private readonly JsonSerializer _jsonSerializer;
+
+        private readonly IServerIdManager _serverIdManager;
+
+        private readonly IMessageBus _messageBus;
+
+        private readonly IEnumerable<IBackplaneMethodResolver> _methodResolvers;
+
+        private readonly ConcurrentDictionary<string, BackplaneChannelRegistration> _subscriptions = new ConcurrentDictionary<string, BackplaneChannelRegistration>();
+
+        public BackplaneChannel(JsonSerializer jsonSerializer, IServerIdManager serverIdManager, IMessageBus messageBus, IEnumerable<IBackplaneMethodResolver> methodResolvers)
+        {
+            _jsonSerializer = jsonSerializer;
+            _serverIdManager = serverIdManager;
+            _messageBus = messageBus;
+            _methodResolvers = methodResolvers;
+        }
+
+        public void Subscribe()
+        {
+            _messageBus.Subscribe(this, cursor: null, callback: HandleMessages, maxMessages: 10, state: null);
+        }
+
+        public void Subscribe<T>(T instance) where T : IDisposable
+        {
+            Subscribe(instance, channelName: null);
+        }
+
+        public void Subscribe<T>(T instance, string channelName) where T : IDisposable
+        {
+            if (string.IsNullOrEmpty(channelName))
+            {
+                channelName = typeof(T).FullName;
+            }
+
+            var channelRegistration = new BackplaneChannelRegistration
+            {
+                Name = channelName,
+                Instance = instance,
+                Type = typeof(T)
+            };
+
+            // overwrite the last registration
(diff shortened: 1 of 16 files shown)


JabbR.Tests/ChatServiceFacts.cs
```diff
@@ -385,7 +385,7 @@ public void ThrowsIfRoomNameIsLobby()
                     Name = "foo"
```
davidfowl, line 385: I think we're going to need to use a DI container in these tests.

JabbR/Infrastructure/IBackplaneChannel.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace JabbR.Infrastructure
+{
+    public interface IBackplaneChannel
+    {
+        void Subscribe<T>(T instance);
```
davidfowl: This method is pretty weird. It should probably also be IDisposable.

JabbR/Infrastructure/IBackplaneChannel.cs
```diff
@@ -0,0 +1,9 @@
...
+
+        void Invoke<T>(string funcName, object[] arguments);
```
davidfowl: methodName

JabbR/Infrastructure/BackplaneChannel.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        private readonly List<string> _signals = new List<string>{ BackplaneSignal };
+
+        private readonly JsonSerializer _jsonSerializer;
+
+        private readonly IServerIdManager _serverIdManager;
+
+        private readonly IMessageBus _messageBus;
+
+        private readonly Dictionary<string, BackplaneChanelRegistration> _subscriptions = new Dictionary<string, BackplaneChanelRegistration>();
```
davidfowl: I think I prefer to mimic the hub proxy model instead of creating flat names, type + method.

JabbR/Infrastructure/BackplaneChannel.cs
```diff
@@ -0,0 +1,145 @@
...
+        {
+            string typeName = typeof(T).FullName;
+            foreach (var method in typeof(T).GetMethods())
+            {
+                _subscriptions.Add(typeName + '+' + method.Name, new BackplaneChanelRegistration
+                    {
+                        Instance = instance,
+                        Method = method 
+                    });
+            }
```
davidfowl, line 80: Make it so you can unsubscribe.

JabbR/Infrastructure/BackplaneChannel.cs
```diff
@@ -0,0 +1,145 @@
...
+                {
+                    // find a subscription
+                    var message = _jsonSerializer.Parse<BackplaneChannelMessage>(m.Value, m.Encoding);
+                    if (_subscriptions.ContainsKey(message.Subscription))
+                    {
+                        // deserialize parameters
+                        var subscription = _subscriptions[message.Subscription];
+                        var parameters = subscription.Method.GetParameters();
+                        var args = new object[parameters.Length];
+                        for (var idx=0; idx<parameters.Length; idx++)
```
davidfowl: Formatting

JabbR/Infrastructure/BackplaneChannel.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        private Task<bool> HandleMessages(MessageResult result, object state)
+        {
+            result.Messages.Enumerate<object>(
+                m => _signals.Any(s => s == m.Key),
+                (s, m) =>
+                {
+                    // find a subscription
+                    var message = _jsonSerializer.Parse<BackplaneChannelMessage>(m.Value, m.Encoding);
+                    if (_subscriptions.ContainsKey(message.Subscription))
```
davidfowl: TryGetValue

JabbR/Infrastructure/BackplaneChannel.cs
```diff
@@ -0,0 +1,145 @@
...
+                            {
+                                args[idx] = _jsonSerializer.Deserialize(paramReader, parameters[idx].ParameterType);
+                            }
+                        }
+                        subscription.Method.Invoke(subscription.Instance, args);
+                    }
+                },
+                state: null);
+
+            var tcs = new TaskCompletionSource<bool>();
```
davidfowl: return Task.FromResult(true);

JabbR/Services/ChatService.cs
```diff
@@ -290,10 +293,11 @@ public class ChatService : IChatService
                                                                                 {"zanzibar","Zanzibar"}
                                                   };
 
-        public ChatService(ICache cache, IRecentMessageCache recentMessageCache, IJabbrRepository repository)
+        public ChatService(ICache cache, IRecentMessageCache recentMessageCache, IJabbrRepository repository, IChatServiceProxy chatServiceProxy)
             : this(cache,
                    recentMessageCache,
-                   repository,  
+                   repository,
+                   chatServiceProxy,
```
davidfowl: Can we hide this behind the ICache?

JabbR.Tests/ChatServiceFacts.cs
cmaish, line 385: I agree, though if we swap to proxying ICache, we don't need to update the tests any more.  Added this as a separate task - https://github.com/JabbR/JabbR/issues/1081

cmaish:
Still need to switch to hub/proxy model rather than methodname based backchannel registration but otherwise should be much improved.

cmaish:
OK, what about this...?

PR #1078, opened 2014-03-28
Configurable ContentProviders

InfinitiesLoop:
Adds two things.
1. The ability to add content providers dynamically through the admin screen. It works for providers that follow a simple pattern: Extract a single piece of data from the URL, and then output HTML using that parameter as an argument. That, or output a script reference that contains document.writes. This allowed me to delete 8 built-in content providers which removes a lot of hard coded urls, patterns, etc, and replace them with 'default' configured providers.
2. The ability to turn off other content providers that are still built-in. Turning a content provider on/off does require restarting the app though, due to the way the application settings object is injected. I could have redone that but it would have been more disruptive than I thought worth doing. A message on the admin page indicates a restart would be required.


davidfowl:
@InfinitiesLoop Would you be willing to make 2. work once this PR is in? It would be great to be able to change content providers without an app restart.

JabbR/ContentProviders/ConfiguredContentProvider.cs
```diff
@@ -0,0 +1,126 @@
...
+        private string WrapWithCapturedWrite(ContentProviderSetting provider)
+        {
+            var script = provider.Script;
+            var title = provider.Title ?? provider.Script;
+
+            var scriptTagId = Guid.NewGuid().ToString();
+            return string.Format(@"
+                    <div id='{0}'></div>
+                    <script type='text/javascript'>
+                             captureDocumentWrite('" + SimpleEscape(script) + @"', '" + SimpleEscape(title) + @"', $('#{0}'));
```
davidfowl, line 113: We've had xsrf issues before so this makes me a little nervous :smile:

davidfowl:
Can you post a screenshot of the admin ui?

davidfowl:
@alfhenrik and @cmaish can you guys take a look as well?

JabbR/ContentProviders/ConfiguredContentProvider.cs
InfinitiesLoop, line 113: It's basically the same thing the old providers were doing, but I shall review again to be sure there's no holes.

InfinitiesLoop:
Admin screen with the new configured content providers. The ones you see are the default providers you will get and correspond to the deleted provider classes in the PR.

image

InfinitiesLoop:
I should mention that the way the UI works is that you can add a new provider by filling in the details on that last, blank row. As soon as you start doing so, it will generate another blank row so you could add another one, etc. Click remove next to each row to delete a provider. It's simple and effective, even if it's not the most beautiful thing in the world. I struggled a bit with how to present the fields in a way that was pleasing but not also noisy. They take up too much space trying to get fancy with them, so I settled on simplicity.

InfinitiesLoop:
Oh and here is the admin UI for enabling/disabling the built-in providers. The name of each is its class name with 'ContentProvider' stripped off the end.

image

PR #1079, opened 2014-03-28
Auto room creation

InfinitiesLoop:
Adds an "auto room creation" admistration option, which makes it so if a user loads jabbr with a room url for a room that does not exist, it is added automatically. This allows for dynamic urls to be used that invite users to discuss a topic without having to make the room in advance. The setting is disabled by default.

Note that this PR builds on top of my previous PR, #1078 because of changes to the administration index.cshtml. Recommend you take that one first if you are going to take both of these.

### DIFF
--- a/JabbR/Views/Administration/index.cshtml
+++ b/JabbR/Views/Administration/index.cshtml
@@ -69,6 +69,13 @@
                         </div>
                     </div>
                     <div class="control-group">
+                        <label class="control-label" for="allowAutoRoomCreation">@LanguageResources.Administration_AllowAutoRoomCreation</label>
+                        <div class="controls">
+                            @Html.CheckBox("allowAutoRoomCreation", applicationSettings.AllowAutoRoomCreation)
+                            <span class="help-block">@LanguageResources.Administration_AllowAutoRoomCreation_Detail</span>
+                        </div>
+                    </div>
+                    <div class="control-group">
                         <label class="control-label" for="maxMessageLength">@LanguageResources.Administration_MaxMessageLength</label>
                         <div class="controls">
                             <input type="text" id="maxMessageLength" name="maxMessageLength" class="input-xlarge" value="@applicationSettings.MaxMessageLength" placeholder="@LanguageResources.Administration_MaxMessageLength"/>
--- a/JabbR/Chat.js
+++ b/JabbR/Chat.js
@@ -308,7 +308,7 @@
     };
 
     // Called when a returning users join chat
-    chat.client.logOn = function (rooms, myRooms, userPreferences) {
+    chat.client.logOn = function (rooms, myRooms, userPreferences, appSettings) {
         privateRooms = myRooms;
 
         var loadRooms = function () {
@@ -357,7 +357,7 @@
         ui.setUnreadNotifications(chat.state.unreadNotifications);
 
         // Process any urls that may contain room names
-        ui.run();
+        ui.run(appSettings);
 
         // Otherwise set the active room
         ui.setActiveRoom(this.state.activeRoom || 'Lobby');
@@ -1090,13 +1090,18 @@
         updateTitle();
     });
 
-    $ui.bind(ui.events.openRoom, function (ev, room) {
+    $ui.bind(ui.events.openRoom, function (ev, room, createIfNotExists) {
         try {
             chat.server.send('/join ' + room, chat.state.activeRoom)
                 .fail(function (e) {
-                    ui.setActiveRoom('Lobby');
-                    if (e.source === 'HubException') {
-                        ui.addErrorToActiveRoom(e.message);
+                    if (createIfNotExists) {
+                        $ui.trigger(ui.events.sendMessage, ["/create " + room, null, true]);
+                    }
+                    else {
+                        ui.setActiveRoom('Lobby');
+                        if (e.source === 'HubException') {
+                            ui.addErrorToActiveRoom(e.message);
+                        }
                     }
                 });
         }
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -1339,13 +1339,12 @@
             }, trimRoomHistoryFrequency);
         },
         
-        run: function () {
+        run: function (appSettings) {
             $.history.init(function (hash) {
                 var roomName = getRoomNameFromHash(hash);
-
                 if (roomName) {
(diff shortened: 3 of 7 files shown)


JabbR/App_Start/Startup.DependencyInjection.cs
```diff
@@ -75,9 +76,6 @@ private static KernelBase SetupNinject(IJabbrConfiguration configuration)
             kernel.Bind<ICryptoService>()
                 .To<CryptoService>();
 
-            kernel.Bind<IResourceProcessor>()
```
davidfowl, line 78: Moved why?

JabbR/Views/Administration/index.cshtml
```diff
@@ -1,8 +1,12 @@
-﻿@using SquishIt.Framework;
+﻿@using System.Collections.Generic
+@using JabbR.ContentProviders.Core
+@using SquishIt.Framework;
 @using JabbR
 @using JabbR.Services
-@model ApplicationSettings
-
+@{
+    var applicationSettings = Model.ApplicationSettings as ApplicationSettings;
```
davidfowl, line 7: Why do you have to cast?

JabbR/Web.config
```diff
@@ -103,6 +103,7 @@
   </entityFramework>
   <razor disableAutoIncludeModelNamespace="false">
     <assemblies>
+      <add assembly="System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089" />
```
davidfowl, line 106: Wut?

davidfowl:
Ahh this was built on top of the other PR. I think this should sit until we review the other one so the diff makes more sense.

JabbR/Views/Administration/index.cshtml
InfinitiesLoop, line 7: Model is dynamic. I generally like to leave the world of dynamics as soon as I can when it makes sense to ensure compile time (or in this case, view-compile-time) checking. The model used to be declared as ApplicationSettings directly, but now it's more of a ViewModel that contains app settings and other things.

JabbR/Web.config
InfinitiesLoop, line 106: System for some reason isn't referenced by default (something with how Nancy is using the view engine?). The codes needed it probably because of the GetType call when determining the name of each provider.

JabbR/App_Start/Startup.DependencyInjection.cs
InfinitiesLoop, line 78: The resource processor now indirectly consumes ISettingsManager (Because the new Configurable Content Provider needs it). But that failed because the settings manager is declared later in the DI setup. Ideally ordering like this doesnt matter with DI but there are some custom injection mappings happening that cause things to resolve before the whole thing is setup. Easiest thing to do was move its declaration so it's after the settings manager declaration.

davidfowl:
@InfinitiesLoop can you resubmit this PR so the diff is friendlier?

InfinitiesLoop:
I forced the PR to update, the diff is much cleaner now. :)

Screenshot of the new admin option

image

InfinitiesLoop:
Any thoughts on this? The PR is older now but it should still be a clean merge.

davidfowl:
This freaks me out a little and I guess since the default is off it's not as bad

InfinitiesLoop:
Yup, off by default. What I love about it is that we can add links to chat rooms for specific, um, 'features' in our internal products. For example, imagine there was a link on this very page that said "Chat about this Pull Request". The link used a room name that included the PR id. The room doesn't exist, but the first time someone uses it, it's created automatically. Thus allowing you to have dynamic chat rooms :) Integration like that just isn't possible with "group lync chat" or what have you.

InfinitiesLoop:
If it would help I could add to the help text on the admin screen to better explain the consequences. Not sure what but I could some up with something.

davidfowl:
Except it's not dynamic. It's in the database forevar.

Buildstarted:
Who would be the creator/owner of the room in the case of auto-creation?

cmaish:
Looks like the person who first visits the new room is the creator as a create command is issued on the client if the join command fails

Buildstarted:
Yeah...I was thinking that's a bad idea...maybe it should be the site admin or some special user? Not sure...just seems odd :)

InfinitiesLoop:
The first user creates the room. It's no different than if the user logs in, doesn't see a chat room they want, and then /creates it. The only difference is they don't necessarily realize they created the room. In my case that's fine, these will typically be rooms that are transient and will die off after a short period of time anyway. It's in the db forever, as you say, but I want that -- the chat history is tied to the original room/topic for future viewers to see, same as this thread is attached to this pull request.

Buildstarted:
That makes more sense now that you spell it out :)

hnrkndrssn:
It seems that as long as the feature is OFF by default, this would be a feature that is useful for custom installs of JabbR (i.e. we probably don't want it on JabbR.net) :grin: 

It's got my :thumbsup:

hnrkndrssn:
@cmaish @davidfowl any thoughts on this one? Do we want to merge it or are there any outstanding questions that needs answering? (I couldn't see any...)

cmaish:
Personal opinions: I'd be for merging it with the default being off, so
long as we have an eventual plan to ship it out to a sattelite assembly
(either under the JabbR project or as a part of another project), possibly
under #1135 or similar.

These sorts of changes are good use cases for JabbR, but they're not really
in what we're doing (or what I see us wanting to do) by default.

Tldr: I haven't been pushing for a merge just yet because I don't see it as
core functionality (and we don't have a good place for this "fringe"
functionality at the moment).  If someone wants or needs it, I'm ok with
shipping and fixing/changing later, but I see it being in the codebase in
its current location only temporarily.

On Thursday, July 24, 2014, Henrik Andersson notifications@github.com
wrote:

> @cmaish https://github.com/cmaish @davidfowl
> https://github.com/davidfowl any thoughts on this one? Do we want to
> merge it or are there any outstanding questions that needs answering? (I
> couldn't see any...)
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/JabbR/JabbR/pull/1079#issuecomment-50112527.

##

InfinitiesLoop:
For what it's worth we are using the setting to great success already :). But if I must use my fork of jabbr to get it, well, there's nothing terribly wrong with that. It might be difficult to keep up with upstream if a lot of changes start happening though.

Issue #1075, opened 2014-03-14
User management in jabbr


0vermind:
Hi guys,
coming from IRC, it seems Jabbr lacks of some must-have user management features.
For example 
1. it's not possible to ban a user who is not connected to the chat room.
2. I cannot find the command to unban a user
3. The is no way to ban users for a certain period of time
4. The is no way yet to block users to register with different user names after the ban (i read some features request about it)

To me this are really important features to move my community from IRC to Jabbr. Please tell me we can unban people :P

0vermind:
This is kinda worrying me.

samandmoore:
you're more than welcome to take a stab at implementing this stuff. open source == 
"show us the codez!11!"

0vermind:
Sorry, to say but this standard reply "it's opensource" does not make any sense to me.
Considering Jabbr is an already published project that seems widely used in asp world, having a _ban_ feature without the ability to _unban_ to me seems like a serious issue. Just like the lack of features I was referring to in my previous post.
My intent is to stimulate a discussion between developers, to drive a re-prioritization of efforts.

samandmoore:
I totally understand your position, and I'm just saying that the best way to "drive a re-prioritization of efforts" is to help with a contribution.

davidfowl:
It's totally fair and I do hope that somebody takes the lead and actually implements these features. There's a bunch of gaps and if you really feel passionate about it, you should help drive it.

davidfowl:
Seems like we have some activity on this! :smile:

PR #1083, opened 2014-03-31
Implemented CheckBanned and Unban (Attempt two)

jonfriesen:
Built two additional commands (unban, checkbanned) and made some changes to
the existing ban command.

**Changes**

ban now shows notification to admin user in bad
custom message if admin tries to ban themselves

**Additions**

unban command: _/unban username_
checkbanned command: _/checkbanned_ or _/checkbanned user_ (shows all banned users or
if a specific users banned status)
I built this PR based on #1075 and it should be noted that the existing
ban command already works on offline users.

I've also added unit tests for both unban and checkbanned commands.

This is my second attempt at posting this. The first time went horrible because I can't GIT. This time I've used an up-to-date branch so my PR shouldn't be stale. Also all commits are squashed instead of a ton pushing.

Hope this is up to your standards, if I can do anything better regarding code style, git pushing, etc I would greatly appreciate the advice!


hnrkndrssn:
Looks good to me

JabbR/Hubs/Chat.cs
```diff
@@ -1199,6 +1199,38 @@ void INotificationService.BanUser(ChatUser targetUser)
...
+        {
+            Clients.Caller.unbanUser(new
+            {
+                Name = targetUser.Name
+            });
+        }
+
+        void INotificationService.IsBanned()
+        {
+            // Get all users that are banned
```
samandmoore, line 1219: is this comment really useful?

samandmoore:
this looks great. it would be swell to also have some tests added for it though: like this. https://github.com/JabbR/JabbR/blob/dev/JabbR.Tests/CommandManagerFacts.cs#L2365

jonfriesen:
@samandmoore, thanks for the advice! Added unit tests for both commands. I'm still grasping how mocking works, so (I think) I'm a little light in that area still.

samandmoore:
lgtm with the tests.

JabbR/Services/INotificationService.cs
```diff
@@ -39,6 +39,9 @@ public interface INotificationService
         void AllowUser(ChatUser targetUser, ChatRoom targetRoom);
         void UnallowUser(ChatUser targetUser, ChatRoom targetRoom);
         void BanUser(ChatUser targetUser);
+        void UnbanUser(ChatUser targetUser);
+        void IsBanned(ChatUser targetUser);
+        void IsBanned();
```
davidfowl: This name needs to change across the board. This should be CheckBanned

PR #1084, opened 2014-04-01
Reload content providers after saving admin

InfinitiesLoop:
Simple change to make it so enabling/disabling a built-in content provider does not require restarting the app.

### DIFF
--- a/JabbR/App_Start/Startup.DependencyInjection.cs
+++ b/JabbR/App_Start/Startup.DependencyInjection.cs
@@ -1,4 +1,7 @@
-﻿using JabbR.ContentProviders;
+﻿using System;
+using System.Linq;
+using System.Reflection;
+using JabbR.ContentProviders;
 using JabbR.ContentProviders.Core;
 using JabbR.Infrastructure;
 using JabbR.Models;
@@ -112,7 +115,9 @@ namespace JabbR
                       .To<SettingsKeyProvider>();
 
             kernel.Bind<IResourceProcessor>()
-                .ToConstant(new ResourceProcessor(kernel));
+                .To<ResourceProcessor>();
+
+            RegisterContentProviders(kernel);
 
             var serializer = JsonSerializer.Create(new JsonSerializerSettings()
             {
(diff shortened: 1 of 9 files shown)


davidfowl:
Can we just make the ResourceProcessor not cache at all?

samandmoore:
@davidfowl you don't think that would hurt performance? it's using MEF to load them up.

InfinitiesLoop:
Yeah I wouldn't want it to reimport with MEF every time. But it could re-evaluate the 'isdisabled' check every time. That wouldn't be so bad. But it would still be a lot of work every request for something that will rarely if ever change at runtime. Up to you...

davidfowl:
Nah, I think we nuke the MEF and just register everything in the ninject

InfinitiesLoop:
It has been updated not to use MEF. Please advise :)

JabbR/App_Start/Startup.DependencyInjection.cs
```diff
@@ -146,5 +151,39 @@ private static KernelBase SetupNinject(IJabbrConfiguration configuration)
 
             return kernel;
         }
+
+        private static void RegisterContentProviders(IKernel kernel)
+        {
+            // Registers content providers by discovering all IContentProvider implementors in all the loaded assembies.
```
davidfowl: Just hard code them.

InfinitiesLoop:
The great advantage this is giving is that it means people can add their own providers without having to modify source code I jabbr. They can just drop in a new dll. We have a few providers we need to add that would use this because they require custom code that the configured provider can't do, and they are specific to our needs so you wouldn't want them in jabbr. Thoughts?

davidfowl:
That was kinda the theory with MEF but it never came to be. We may as well just hard code it for now.

cmaish:
Over a future PR, I'd like to see https://github.com/JabbR/JabbR/blob/master/JabbR/App_Start/Startup.DependencyInjection.cs#L145 get a kernel.Load("*.plugin.dll") call (alongside hardcoded "core" IContentProviders).

We'd have the benefits of not using reflection to discover all IContentProviders (so a hardcoded list of content providers would work), not needing to mess with <assembly> attributes and still being able to drop in myJabbRPlugin.plugin.dll files to extend JabbR's behavior (in addition to adding content providers, we'd be able to override other services).  I'm happy to look at this (and some associated refactorings) sometime soonish if it'd help.

InfinitiesLoop:
I see. We have the need for some custom providers, so hope to see this supported somehow. In the meantime I'll update it to be hard coded, and we can continue the discussion via another pr :)

cmaish:
Makes sense.  This morning I threw together a quick PoC of what I was talking about, and while fragile, it works pretty well.  I'm definitely going to harass @davidfowl about getting plugin support in the codebase because I think it'd increase the number of contributions (...and make it so that people can build addons/features without needing to fork and keep updating their own branch of JabbR).

davidfowl:
Whats the status of this thing? Can we hard code for now then figure out plugins?

InfinitiesLoop:
It is now hard coded! And I've done a rebase/squash on my changes so the plugin business is like it never happened.

davidfowl:
@InfinitiesLoop hiding the shame? :trollface:

davidfowl:
This looks good

InfinitiesLoop:
What shame? #seewhatididthere

Issue #1055, opened 2014-01-22
Download messages is disabled when room is private but no notification


Buildstarted:
When a room is private the "Download messages" dialog is disabled and doesn't show. However there is no visual notification that they're disabled because it's a private room. 

![image](https://f.cloud.github.com/assets/743025/1971340/05cf5760-8329-11e3-95ac-071739859cdb.png)

The option could be grayed out when switched to a private room or some other visual indication that the room doesn't allow messages to be downloaded.

cmaish:
I think that we should add a modal dialog or something if you do try, so that it's clear that it can't be done.

PR #1090, opened 2014-04-12
Added toggle for download messages

Buildstarted:
When room is locked there is no UI to indicate that
you cannot download messages #1055
- Added new download icons.
- Added toggleDownloadButton method
- Updated methods to call when changing rooms and
  on locking room


Buildstarted:
fixing commits

PR #1091, opened 2014-04-12
Added toggle for download messages/paste images additions

Buildstarted:
When room is locked there is no UI to indicate that you cannot download
messages #1055
- Added new download icons
- Added toggleDownloadButton to Chat.ui.js
- Updated room methods to change button status on room change and lock
  change
- Updated paste dialog with new keyhandlers for <esc> and <enter>
- Focus message box after dialog closed.


davidfowl:
Where did the icons come from?

Buildstarted:
From the `glyphicons-halflings.png` included in JabbR. I just copied the cancel icon on top of the download icon to get the same look as the richness status.

xt0rted:
Since Font Awesome is also used why not use stacked icons instead? http://fortawesome.github.io/Font-Awesome/examples/#stacked

Buildstarted:
Because I used the same structure as the other disabled icon. :) I don't see a problem off-hand with using it.

PR #1096, opened 2014-05-11
Fix for Splash screen fails closing

### LINKED ISSUE #1095: Load animation causing login to hang/lock
The "R" circle animation appears to be causing some users to "hang" after logging in and it prevents the remaining parts of the page to load. This is cross browser (IE8, FF). Can something be done to easily remove or turn off that feature?

hnrkndrssn:
Fixes #1095

### DIFF
--- a/JabbR/Chat.js
+++ b/JabbR/Chat.js
@@ -92,9 +92,14 @@
             })
             .fail(function (e) {
                 connection.hub.log('loadRooms.failed(' + rooms.join(', ') + ', ' + e + ')');
+            })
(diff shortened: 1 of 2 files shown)


InfinitiesLoop:
Ah ha, we had a user affected by this and I couldn't figure it out. We had to nuke their room membership in the db. Please merge! :)

davidfowl:
Yah, 5 times seems like enough :smile:

davidfowl:
:shipit:

PR #1102, opened 2014-05-19
Fix /list users

### LINKED ISSUE #1101: /list is a little broken
Seems that a while back, I refactored some code underlying the list command from using $.each to $.map, and $.map doesn't use function.call, so we get the following:
capture 9

What it should look like is a list of user names

cmaish:
Fixes #1101 /list is a little broken.  Also removes code that is not used from the very-similar and now-removed room listing command.


davidfowl:
:shipit:

PR #1106, opened 2014-05-19
Add notification.mp3 to support IE

garysharp:
Internet Explorer [doesn't support](http://en.wikipedia.org/wiki/Html5_audio#Supported_audio_codecs) the WAV PCM format in `audio` tags. This PR contains `notification.mp3` (which was converted from `notification.wav`) and updates the `#notificationSound` `audio` tag to use this format if the browser prefers.


hnrkndrssn:
According to this page http://www.w3schools.com/tags/tag_audio.asp the mime type for mp3 audio is 'audio/mpeg', will 'audio/mp3' still work?

garysharp:
> the mime type for mp3 audio is 'audio/mpeg', will 'audio/mp3' still work?

Looks to be some conflicting advise on this throughout the internet. VS tooling (and elsewhere) suggested `audio/mp3`, various other sources suggest `audio/mpeg` as you say.

Also conflicting advise on PCM WAVE format, some say `audio/wav`, others `audio/wave`. Mozilla [weighs in](https://developer.mozilla.org/en-US/docs/HTML/Supported_media_formats) by _preferring_ `audio/wave` - perhaps this is the way to go?

Testing revealed that both mime-types worked in IE 11 (uses mp3), Chrome 34 (uses wav), and Firefox 29 (uses wav).

garysharp:
Updated mime-types:

> `audio/mp3` -> `audio/mpeg`
> `audio/wav` -> `audio/wave`

cmaish:
Looks and sounds good to me :shipit: IMO

davidfowl:
:shipit:

PR #1112, opened 2014-05-20
Improve perf user status change

cmaish:
Figured that I should at least make this into a PR.  What this does is only performs status updates to inactive only if the user was not previously marked as inactive.  This is to try to mitigate performance issues with the inactive updater process.

Might be superseded by other work on playing around with CSS rules directly.


hnrkndrssn:
Looks good to me :shipit:

PR #1115, opened 2014-05-22
Custom error page

cmaish:
This one's been sitting around for a while.  

Add an error page, make it suggest rooms if you try to access site.com/roomName and it's a 404.


xt0rted:
Since you can link to rooms like so `https://jabbr.net/#/rooms/meta` what about handling the format `https://jabbr.net/rooms/meta` as well?

cmaish:
Should be ready to go now.

hnrkndrssn:
Looks good to me :+1:

cmaish:
Should be good to merge on the new dev!

davidfowl:
pics of the error page

cmaish:
croppercapture 4

The suggestion line only shows when you've entered a room-like url.  The 404 - NotFound is the code and Enum value that nancy gives us.

Issue #1047, opened 2013-11-07
Kicking is broken after 1st kick


Yantrio:
1. kick person,
2. person re-joins
3. kick person again (fails)

we would love to fix this asap :)

adamralph:
+1 loads of spammers on recently

cmaish:
This is related to the user-in-room cache which is maintained on each WFE individually.  We need some sort of communication between servers to handle cache invalidation... if only there were some real-time cross-machine communication library that we could use.

cmaish:
Note this is also the cause of users sometimes not appearing in the userlist correctly (ie appearing as offline etc)

PR #1117, opened 2014-05-26
Mitigate issues with room cache and scaleout

cmaish:
#1047 is an issue because we have an effectively infinite cache on the front end servers.  Let's just make the timeout really low.

This change will have a negative impact on server performance.


davidfowl:
:shipit: , lets see how it affects performance. We may want to be a little smarter and make the cache keep entries while it's hot.

PR #1125, opened 2014-05-30
Markdown subset support

### LINKED ISSUE #1124: Basic markdown support for messages
Allow some basic markdown support for things like **bold** _italic_ and `inline code`

khellang:
Fixes #1124

This PR replaces StackExchange's PageDown with [marked](https://github.com/chjj/marked) which supports [GitHub Flavored Markdown](http://github.github.com/github-flavored-markdown) (GFM). It also adds highlight.js for syntax highlighting of most languages.

GitHub issues, PRs and comments all use the full GFM parser (block and inline elements), while content (chat messages, topic, etc.) only allows inline elements, like `code`, `_italics_` and `**bold**`. This makes sense while there still isn't support for multiline chat messages. Pasted text is not parsed for markdown. I've disabled markdown links, so you're not able to do `[google.com](some.malicious-site.com)` :smile:

Both chat and GitHub issues/comments parsers use sanitation, so it should be safe from XSS.

You can test this out on http://jabbr-markdown.azurewebsites.net/


JabbR/Content/themes/default/Chat.githubissues.css
```diff
@@ -117,7 +117,7 @@
 .new-comments .repo-owner-tag .tag, .new-comments .gist-owner-tag .tag {
     background: none repeat scroll 0 0 #2CC03E;
     border-color: #259A33 #2CC03E #2CC03E #259A33;
-}
+}*/
```
xt0rted: Should this be removed instead?

khellang: Yeah. I wasn't sure if it was used anywhere (I couldn't find anything), but it was screwing up the syntax highlighting (it uses `.tag`), so I just commented it out. I'll just remove it for now...

davidfowl:
@khellang whats up with code blocks?

khellang:
What's up with it?

Code blocks

davidfowl:
@khellang 2 things:
- We need a way to support multiline code
- There's a bug with encoding in code blocks

Try typing

```
`if (x < 5)`
```

khellang:
I'll look into the encoding issue, I think we can just drop the explicit encoding and let the markdown parser fix it. Do you feel multiline code should be part of this PR? I figured it was best to start out with single-line, since we don't support multiline input yet (it's treated as "pasted").

davidfowl:
Yea lets to multiline in another PR

khellang:
OK, I removed the extra encoding and pushed the new version. Want to see if you can break it now? :wink:

PR #1126, opened 2014-06-02
Fix csrf validation

### LINKED ISSUE #1093: CSRF token not validated
Signin page (among others) has a CSRF token embedded in the form (Html.AntiForgeryToken) but it's never checked.  We should probably call ValidateCsrfToken somewhere.

cmaish:
Adds validation for CSRF tokens.  Fixes #1093

When combined with #1115, looks like:

capture 25

### DIFF
--- a/JabbR.Client.Sample/Program.cs
+++ b/JabbR.Client.Sample/Program.cs
@@ -166,6 +166,7 @@ namespace JabbR.Client.Sample
             {
                 BaseAddress = new Uri(server)
             };
+            client.DefaultRequestHeaders.Add("sec-jabbr-client", "1");
 
             var content = new FormUrlEncodedContent(new Dictionary<string, string>
             {
--- a/JabbR.Client/DefaultAuthenticationProvider.cs
+++ b/JabbR.Client/DefaultAuthenticationProvider.cs
@@ -35,6 +35,7 @@ namespace JabbR.Client
             };
 
             var client = new HttpClient(handler);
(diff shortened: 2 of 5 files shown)


hnrkndrssn:
:thumbsup: looks good to me

davidfowl:
@cmaish Check JabbR.Client

cmaish:
I'm thinking that we can solve that by checking for a Sec- header
(Sec-Jabbr-Client?), which, if present will skip the CSRF token check.
 This way, we've got protection against ajax-based posting, but it's pretty
straightforward to use non-browser-based clients to connect to jabbr.

Will update to show what I mean when I get the chance.

davidfowl:
@cmaish yep

cmaish:
Should be ready to go now.  _Note that this does break existing clients_ - they need to send the sec-jabbr-client header (or a valid CSRF token) or they'll get errors.

davidfowl:
This will break older clients

cmaish:
useragents containing "signalr" and also useragents which do not report headers are now exempt from the CSRF check.  Newer agents could (and should) still use the sec-jabbr-client header, but this allows for older agents to not be broken.

cmaish:
Just in case a client cannot be upgraded, I've added an admin option that allows the user to disable the CSRF check at a server level.
capture 40

cmaish:
Reverted the change for the admin option, we should have CSRF always validated and just tell people to update their bots.

PR #1127, opened 2014-06-11
Add french language resources

tbolon:
Here is a pull request to use the LanguagesResources feature to add support for the french language.

Note that there are still issues where text is not localized. But I am not sure how the language to use should be determined client-side (server globalization settings, or client browser language ?). It works for most items (views) but not for strings returned from hub (command help for instance).


cmaish:
Thanks for the translations!  I'd like to get them in as soon as possible but they do reveal some problems with how we do our localisation (things like server responses sometimes not being properly localised require semimajor changes), which mean that it might be a little bit of time before these changes appear in a dev build.

Issue #1131, opened 2014-07-02
Notifications page is busted


cmaish:
The notifications page does not work at current.

The cause for this is most likely a 404 on the marked.js script on the notifications page.  In dev this is a 404, in prod the bundling seems to throw an exception.

hnrkndrssn:
Fixed in #1132

PR #1132, opened 2014-07-02
Fix notifications page's 404, #1131

cmaish:


hnrkndrssn:
Looks good to me :shipit:

PR #1138, opened 2014-07-05
Add zero-width space character before and after matches for emoji aliases

### LINKED ISSUE #1054: Character incorrectly identified as part of emoticon
If you type ":)p" in the chat window, it is converted to the smiling emoticon with "tongue:" immediately following it (note the lack of colon on the front of tongue:). 

In this case I would expect to see the smiling emoticon followed by a p but it seems to be associating the p with the initial colon.

cmaish:
make it so that :)P doesn't evaluate to :smile:P (which then evaluates to :smile:tongue:, which gets shortened to :smile: tongue:).

Fixes #1054

Insert zero-width spaces before and after substituted tokens.
croppercapture 5

### DIFF
--- a/JabbR/Chat.emoji.js
+++ b/JabbR/Chat.emoji.js
@@ -27,7 +27,7 @@
             for (var key in validAlias) {
                 if (validAlias.hasOwnProperty(key)) {
                     var regex = new RegExp(key, "g");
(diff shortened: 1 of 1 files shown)


davidfowl:
:shipit:  Clever

Issue #1116, opened 2014-05-23
Cancel Google OAuth login shows wrong error message


stimpy77:
If you cancel logging in via Google OAuth, the error message is "Index (zero based) must be greater than or equal to zero and less than the size of the argument list."

cmaish:
This looks like it's an issue with SimpleAuthentication (which isn't fixed by updating to the newest version).  When google authentication is backed-out-of, IAuthenticationCallbackProvider's Process method is called with this error.

hnrkndrssn:
Ping @PureKrome

Sent from my Windows Phone

cmaish:
https://github.com/SimpleAuthentication/SimpleAuthentication/issues/146 created

cmaish:
https://github.com/SimpleAuthentication/SimpleAuthentication/pull/147 created, just need to wait for the SimpleAuthentication change to get into a build to pull it into JabbR.

PureKrome:
Ok. new packages are up on NuGet.org.

You should be updating to [v SimpleAuthentication 0.3.14](http://www.nuget.org/packages/SimpleAuthentication.Core/).

Issue #963, opened 2013-07-29
Error on adding Google accounts where the name contains a hyphen


fatso83:
There seems to be some kind of parsing error when constructing user names from real names. My registered given name in my google account is "Carl-Erik". This seems to have been converted to "Carl Erik", which apparently is illegal.

This all happened when I was trying to create a new account and pressed the Google button/link on the sign-in form.

![screen shot 2013-07-29 at 10 42 21](https://f.cloud.github.com/assets/618076/870380/5c48d49c-f82b-11e2-93d8-be1912f38b3c.png)

chrisfcarroll:
jabbr.net still has this error as of today. I get "[InvalidOperationException: Chris F is not a valid user name.]" when attempting to sign up with google

cmaish:
I don't think that a new deployment has been done just yet, so while this should be fixed when new builds are deployed, it'll probably take a couple of weeks before davidfowl gets around to pushing this change out.

chrisfcarroll:
Noted, thanks.

PR #1139, opened 2014-07-05
Fix google authentication back-out message (#1116)

cmaish:
croppercapture 3

### DIFF
--- a/JabbR/Middleware/WindowsPrincipalHandler.cs
+++ b/JabbR/Middleware/WindowsPrincipalHandler.cs
@@ -30,11 +30,8 @@ namespace JabbR.Middleware
 
                 if (context.Response.StatusCode == 401)
                 {
-                    // We're going no add the identifier claim
-                    var nameClaim = windowsPrincipal.FindFirst(ClaimTypes.Name);
-
-                    // This is the domain name
-                    string name = nameClaim.Value;
+                    // We're going to add the identifier claim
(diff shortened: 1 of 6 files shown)


cmaish:
#963 actually seems to be an issue with spaces in the first/last name on the Google account (or possibly special dashes - we do allow regular dashes and these seem to work).  Regardless now, anything that makes it an invalid user name should be removed.

JabbR/Middleware/WindowsPrincipalHandler.cs
```diff
@@ -30,11 +30,8 @@ public WindowsPrincipalHandler(AppFunc next)
 
                 if (context.Response.StatusCode == 401)
                 {
-                    // We're going no add the identifier claim
-                    var nameClaim = windowsPrincipal.FindFirst(ClaimTypes.Name);
-
-                    // This is the domain name
-                    string name = nameClaim.Value;
+                    // We're going to add the identifier claim
+                    string name = windowsPrincipal.GetUsername();
```
davidfowl: ?

cmaish: It does pretty much the same thing, just using the extension method in https://github.com/JabbR/JabbR/pull/1139/files#diff-6f816fea17915437125c0374c9213f16R58 which additionally ensures that the username contains only jabbr-safe characters and is 30 chars or less

PR #1140, opened 2014-07-05
Fix #1094

### LINKED ISSUE #1094: Cannot disable all content providers
At least one content provider must be enabled.

https://github.com/JabbR/JabbR/blob/dev/JabbR/Nancy/AdministrationModule.cs#L62 's has EnabledContentProviders == null in this case, so we get an exception.

.Where(typeName => enabledContentProvidersResult.EnabledContentProviders == null || !enabledContentProvidersResult.EnabledContentProviders.Contains(typeName)) should work...

cmaish:
Make it so that all content providers can be disabled without an error.

### DIFF
--- a/JabbR/Nancy/AdministrationModule.cs
+++ b/JabbR/Nancy/AdministrationModule.cs
@@ -59,7 +59,8 @@ namespace JabbR.Nancy
                     settings.DisabledContentProviders =
                         new HashSet<string>(contentProviders
                             .Select(cp => cp.GetType().Name)
-                            .Where(typeName => !enabledContentProvidersResult.EnabledContentProviders.Contains(typeName))
+                            .Where(typeName => enabledContentProvidersResult.EnabledContentProviders == null ||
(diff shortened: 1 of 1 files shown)


davidfowl:
:shipit:

jrmcdona:
I have downloaded the zip file and ran the project locally in Visual Studio.  What is the best practice for getting these fixes into Visual studio which isnt hooked up through Git.   I was planning to run the project out of Visual Studio at work as we do with other projects.  BUt would like to get these fixes that are going in such as the notifications fix.   Sorry, it is a GIT newbie question.   Thanks Jordan

cmaish:
The dev archive should have the fixes that have been merged into dev in the source code (https://github.com/JabbR/JabbR/archive/dev.zip), but best practice (and the easiest way to get PRs that haven't yet been accepted) is really to get git up and running - you don't _need_ to hook it into your IDE, so it shouldn't interfere with your existing SCC provider.

PR #1142, opened 2014-07-05
fix #1008

### LINKED ISSUE #1008: When the lobby is focussed, the search rather than send textbox should probably be selected.
At the moment, when you focus on the lobby, the message send textbox is focussed.  This should probably be the room filter search textbox.

cmaish:
Focus the search textbox when the lobby is selected, otherwise focus on the chat message box.


hnrkndrssn:
Looks good! :shipit:

PR #1143, opened 2014-07-06
Improve kick

cmaish:
Kick, ban etc are now all slightly nicer.  If you're kicked, you get told in a popup notification.  If you're banned, same.  Reasons are supported for both kick and ban.  Users in rooms get notified of other users being kicked/banned (and by who, and if there's a reason, what that is)
croppercapture 6
croppercapture 7
croppercapture 8


hnrkndrssn:
Looks alright to me!

jrmcdona:
Should this be in the dev archive zip?   I don't see the chat.client.ban function update in it.  Nor do I see it when I cloned today.   When I ran  git pull https://github.com/JabbR/JabbR improve-kick it tells me already up to date.  However, the Chat.js file still isnt showing chat.client.ban function update that comes along with this PR.   Thank you!

hnrkndrssn:
This has not yet been merged into dev, so no it won't be in the dev archive.

jrmcdona:
Chris are you planning to push to dev?  Thanks

Buildstarted:
@cmaish i think reason should be required - just so that there's accountability

cmaish:
Requires a reason for kick/ban could be something that's configurable, though as a general rule I don't like anything mandatory unless it's required for some system functionality.  Maybe we could get this in and also raise a bug (feature req.) for togglable mandatory-ness.

hnrkndrssn:
:+1: Get it in and then we can enhance it!

PR #1144, opened 2014-07-06
[Prereq #1143] Cant download messages dialog

cmaish:
Push #1143 first.

Add a dialog which says "no, you can't download the messages from this room."

croppercapture 10


hnrkndrssn:
The changes for blocking downloading of messages in the lobby and a closed room look fine. :shipit: (once #1143 has been merged :grin:)

PR #1146, opened 2014-07-09
Gravatar hover preview

### LINKED ISSUE #516: Gravatar hover preview effect
Hi Mr Fowler,

I think a nice little feature that could easily be added to the UI, is a hover preview of a users Gravatar image.

Basic coding which works is;

$('.gravatar').hover(function(){  
    var $g = $(this);
    $g.attr('src', $g.attr('src').replace('s=16','s=80'));
 }, 
function(){ 
    var $g = $(this);
    $g.attr('src', $g.attr('src').replace('s=80','s=16')); 
});

hnrkndrssn:
This PR is to get an MVP for this feature out there, then we can expand it further as we want to.

User list hover:

Message list hover:

Fixes #516

### DIFF
--- a/JabbR/Chat.ui.js
+++ b/JabbR/Chat.ui.js
@@ -78,10 +78,7 @@
         roomLoadingTimeout = null,
         Room = chat.Room,
         $unreadNotificationCount = null,
-        $splashScreen = null,
-        gravatarPopover = null,
-        gravatarCard = null,
-        largeGravatar = null;
+        $splashScreen = null;
 
     function getRoomNameFromHash(hash) {
         if (hash.length && hash[0] === '/') {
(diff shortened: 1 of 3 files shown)


cmaish:
In firefox, image is pushed to a newline after the status bar.

Popups sit within their parent causing issues sometimes (like if you hover over the last user's name in the chat window, it adds scroll height).  Ideally we'd detect if the image can't go below and put it above, but I'd be equally OK with it just overflowing somewhere that gets hidden at the moment.

cmaish:
Also, name is hidden in the popup when the screen is narrow and it's hidden in the chat room (there's a media query rule for @media screen and (max-width:480px){ .messages .message .left .name { display:none; } })

hnrkndrssn:
Yep, I'll take a look at those...

hnrkndrssn:
Updated hover in the messages list

image

cmaish:
There's still some (unavoidable?) overflow when you're scrolled to the bottom of a chat window and hover over a user's name, but IMO this is fine.

When you have narrow screen widths, the userlist image gets pushed down (in FF and Chrome at least):

image

hnrkndrssn:
Alright, I've made some further updates (made the preview smaller when on small screens) that should hopefully sort this out. I've tested the changes in Chrome, FF, and IE and it looks good to me. :smile:

cmaish:
Looks good to me, IMO :shipit:

PR #1147, opened 2014-07-09
Add /lmgtfy and /lmbtfy commands

garysharp:
`/lmgtfy SignalR Documentation` becomes:
lmstfyscreenshot

### DIFF
--- a/JabbR/Commands/LmbtfyCommand.cs
+++ b/JabbR/Commands/LmbtfyCommand.cs
@@ -1,0 +1,31 @@
+﻿using JabbR.Models;
+using Microsoft.AspNet.SignalR;
+using System;
+
+namespace JabbR.Commands
+{
+    [Command("lmbtfy", "Lmbtfy_CommandInfo", "query", "user")]
+    public class LmbtfyCommand : UserCommand
+    {
+        public override void Execute(CommandContext context, CallerContext callerContext, Models.ChatUser callingUser, string[] args)
+        {
+            if (String.IsNullOrEmpty(callerContext.RoomName))
+            {
+                throw new HubException(LanguageResources.InvokeFromRoomRequired);
+            }
+
+            if (args == null || args.Length < 1)
+            {
+                throw new HubException(LanguageResources.Lmstfy_DataRequired);
+            }
+
+            ChatRoom callingRoom = context.Repository.GetRoomByName(callerContext.RoomName);
+
+            string query = string.Join(" ", args);
+            string queryEscaped = Uri.EscapeDataString(query);
+            string url = String.Format("http://letmebingthatforyou.com/?q={0}", queryEscaped);
+
+            context.NotificationService.SendMessage(callingRoom, url);
+        }
+    }
+}
--- a/JabbR/Commands/LmbtfyCommand.cs
+++ b/JabbR/Commands/LmbtfyCommand.cs
@@ -1,10 +1,13 @@
-﻿using JabbR.Models;
+﻿using JabbR.Infrastructure;
+using JabbR.Models;
 using Microsoft.AspNet.SignalR;
 using System;
+using System.IO;
+using System.Linq;
 
 namespace JabbR.Commands
 {
-    [Command("lmbtfy", "Lmbtfy_CommandInfo", "query", "user")]
+    [Command("lmbtfy", "Lmbtfy_CommandInfo", "[@user] query", "user")]
     public class LmbtfyCommand : UserCommand
     {
         public override void Execute(CommandContext context, CallerContext callerContext, Models.ChatUser callingUser, string[] args)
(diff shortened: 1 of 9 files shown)


garysharp:
Rebased this PR (to include #1149) as #1148 made it hard to evaluate the `lmgtfy` and `lmbtfy` commands.

JabbR/LanguageResources.resx
```diff
@@ -1245,4 +1245,19 @@ Use ? or type /? to display the FAQ and list of commands.</value>
...
+  </data>
+  <data name="Lmgtfy_DataRequired" xml:space="preserve">
+    <value>A search query is required.</value>
+  </data>
+  <data name="Lmbtfy_DataRequired" xml:space="preserve">
+    <value>A search query is required.</value>
+  </data>
+  <data name="Lmgtfy_DataRequired" xml:space="preserve">
+    <value>A search query is required.</value>
+  </data>
```
hnrkndrssn, line 1292: These look to be duplicated....?

JabbR/LanguageResources.resx
```diff
@@ -1245,4 +1245,19 @@ Use ? or type /? to display the FAQ and list of commands.</value>
   <data name="Administration_AppInsightsId" xml:space="preserve">
     <value>AppInsights ID</value>
   </data>
+  <data name="Lmbtfy_CommandInfo" xml:space="preserve">
+    <value>Posts a link instructing how a user can Bing the query for themselves using the lmbtfy.com API.</value>
+  </data>
```
hnrkndrssn, line 1256: Shouldn't there be one for Lmgtfy as well?

JabbR/Commands/LmbtfyCommand.cs
```diff
@@ -0,0 +1,31 @@
...
+                throw new HubException(LanguageResources.Lmbtfy_DataRequired);
+            }
+
+            ChatRoom callingRoom = context.Repository.GetRoomByName(callerContext.RoomName);
+
+            string query = string.Join(" ", args);
+            string queryEscaped = Uri.EscapeDataString(query);
+            string url = String.Format("http://lmbtfy.com/?q={0}", queryEscaped);
+
+            context.NotificationService.PostNotification(callingRoom, callingUser, url);
```
cmaish: Why is this postnotification while LMGTFY uses sendmessage?  You don't get a sender username when you call this.

JabbR/LanguageResources.resx
cmaish, line 1256: In help, there are notes for /lmbtfy, but nothing for /lmgtfy:

image

JabbR/LanguageResources.resx
cmaish, line 1292: Your call: maybe share the "A search query is required" resource text between both lmgtfy and lmbtfy when fixing this up.

JabbR/LanguageResources.resx
garysharp, line 1256: _Rebase fail..._ Fixed, its there now.

JabbR/Commands/LmbtfyCommand.cs
garysharp: This was clearly a bug (mine).
I have changed both to `SendMessage`. I've also removed the `user` parameter from this method as its not used anywhere (add confusion as this `user` parameter isn't used to send the message as assumed).
This is building (and generalising) on what _was_ the `GenerateMeme` method ([introduced here](https://github.com/JabbR/JabbR/commit/a8e4db6feeabe1cda7e859a66fa49004cd51f129#diff-929c50cc9143b4a6b528e721c2c1a1b6R32)).

JabbR/LanguageResources.resx
garysharp, line 1292: Removed the duplicate entries (_rebase fail_) - and I combined the `Lmbtfy_DataRequired` (Bing) and `Lmgtfy_DataRequired` (Google) entries into one: `Lmstfy_DataRequired` (s = Search)

garysharp:
It would seem that http://letmebingthatforyou.com/ might be a better option than http://lmbtfy.com/
The former has no advertising and more closely resembles Bing. However, the URL itself is more obvious.

I've updated this PR to reflect this - let me know if you want it changed back.

**Ideas:**
- URL shortening service
- `@username` parameter to alert a particular user
  _For example:_ `/lmgtfy @gsharp JabbR` -> `@gsharp http://lmgtfy.com/?q=JabbR`

garysharp:
URL shortening (tinyurl) has been implemented. JabbR Commands aren't async, so this implementation is synchronous.

User notification has also been implemented as described above.

JabbR/Commands/LmbtfyCommand.cs
```diff
@@ -0,0 +1,65 @@
...
+            }
+
+            if (args == null || args.Length < 1)
+            {
+                throw new HubException(LanguageResources.Lmstfy_DataRequired);
+            }
+
+            string query = null;
+            ChatUser alertUser = null;
+            if (args[0].StartsWith("@"))
```
cmaish, line 27: I'm on the fence about this bit and want to wait for a decision from davidfowl.  We might wind up switching this to being mandatory and not requiring an @ (which is our convention elsewhere)...

garysharp, line 27: As per discussion in JabbR's meta room happy to change implementation to whatever the masses (or our benevolent dictator) decide.

PR #1149, opened 2014-07-09
disable link recognition in the markdown parser.  This prevents links with schemes from being processed twice (fixes #1148)

### LINKED ISSUE #1148: Linking broken in dev when using scheme
Enabling the gfm flag on marked.js means that links are being autolinked and so appearing twice:

clipboard_95b6

cmaish:


davidfowl:
:ship it:

davidfowl:
:shipit:

PR #1153, opened 2014-07-13
Make our command parser less eager

### LINKED ISSUE #1152: Command parsing is too eager
image

we should probably only attempt to execute a command if the message starts with `/` and immediately is followed by [a-zA-Z]

hnrkndrssn:
Fixes #1152

### DIFF
--- a/JabbR/Commands/CommandManager.cs
+++ b/JabbR/Commands/CommandManager.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.ComponentModel.Composition.Hosting;
 using System.Linq;
+using System.Text.RegularExpressions;
 using JabbR.Models;
 using JabbR.Services;
 using Microsoft.AspNet.SignalR;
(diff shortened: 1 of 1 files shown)


cmaish:
I'm not a big fan of how execution flows around in CommandManager - Regex should probably be something like "^\/[A-Za-z0-9]+?".  I don't mind including unicode chars (will unicode chars in command names ever be supported?).

My issue with this change/the existing is that if you enter:

"a /b" it'll go through and match, then fall into https://github.com/JabbR/JabbR/blob/master/JabbR/Commands/CommandManager.cs#L89
"//b" it'll go through and match, and do the same
"/:" will give an invalid command error (not a problem, but it'd be nice if it sent through anything that didn't look like it was deliberately a comment)

It'd be nice to roll the two TryHandleCommand methods together, but IMO that's probably out of scope for a fix this small.

cmaish:
Looks good to me, I'd :shipit:

PR #1156, opened 2014-07-15
XKCD Content Provider support (take 2)

ctolkien:
Heavily based off the existing Bash content provider, although no longer quite so much @khellang :)

Looks like this

Has alt text as the title of the image, so no spoilers.

### DIFF
--- a/JabbR/ContentProviders/XkcdContentProvider.cs
+++ b/JabbR/ContentProviders/XkcdContentProvider.cs
@@ -67,7 +67,7 @@ namespace JabbR.ContentProviders
 
         public override bool IsValidContent(Uri uri)
         {
-            return uri.AbsoluteUri.Matches(@"https?:\/\/xkcd.com\/\d+\/");
+            return uri.AbsoluteUri.Matches(@"https?:\/\/xkcd.com\/\d+\/?");
         }
 
 
--- a/JabbR/ContentProviders/XkcdContentProvider.cs
+++ b/JabbR/ContentProviders/XkcdContentProvider.cs
@@ -15,7 +15,7 @@ namespace JabbR.ContentProviders
         private static readonly string ContentFormat = @"
 <div class=""xkcd_wrapper"">
     <div class=""xkcd_header"">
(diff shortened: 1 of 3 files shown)


JabbR/ContentProviders/XkcdContentProvider.cs
```diff
@@ -0,0 +1,82 @@
...
+            return ExtractFromResponse(request).Then(pageInfo =>
+            {
+                if (pageInfo == null)
+                {
+                    return null;
+                }
+
+                return new ContentProviderResult
+                {
+                    Content = String.Format(ContentFormat, request.RequestUri.ToString(),  pageInfo.ImageUrl, pageInfo.Title, pageInfo.Description),
```
hnrkndrssn: You're passing 4 parameters, but your only using 3(?) of them in the actual `ContentFormat` string...is there are reason for this?

hnrkndrssn:
This works:
http://xkcd.com/13/

This does not:
http://xkcd.com/13

image

Should we make it check, and append a '/', if the URL doesn't have it?

JabbR/ContentProviders/XkcdContentProvider.cs
ctolkien: Good catch, this is a holdover from adjusting what the output looked like.

ctolkien:
@alfhenrik  I enforced the ending trail slash as a.) the website 301's to having the trailing slash (so if you copy pasta from the website, it will have it) and I wasn't sure if the 301 would be followed.

hnrkndrssn:
Yea, I guess you'd be copying it from the website, not manually typing it in like I did :grin:

hnrkndrssn:
Looks good to me :+1: 

@davidfowl @cmaish wanna chime in?

khellang:
It would be nice to allow urls without the trailing slash, after all, xkcd.com allows it :smile:

ctolkien:
Now works with or w/o slash. Seems redirects are followed, so we don't need to append it at our end. 

image

khellang:
:metal:

PR #1160, opened 2014-07-17
Resend failed msg polish

### LINKED ISSUE #1159: Resend failed message improvements
Currently if you send `/foo` the resend icon will show and allow you to resend the message, this should not happen.

If you are not connected the resend function should not be available.

hnrkndrssn:
We should not display the resend button for failed commands or if we are disconnected.

Fix #1159


cmaish:
Looks great to me.

PR #1161, opened 2014-07-18
Display a users owned rooms on their profile page.

### LINKED ISSUE #718: Show rooms owned by user on account page
When I go to my account page it would be nice to have a list of rooms I own listed.

hnrkndrssn:
Will show as:
image

The rooms are not links that you can click to open the room, I thought as this would open a new tab (i.e. not in the browser tab that JabbR is currently open in) you would end up with a number of tabs open.

But if others think that the rooms should be links opening the room, then I'll make the change.

 Fixes #718

### DIFF
--- a/JabbR/Content/themes/default/Chat.css
+++ b/JabbR/Content/themes/default/Chat.css
@@ -372,6 +372,8 @@ canvas {
 
 #owned-rooms li {
     cursor: default;
+    width: 240px;
+    margin: 0 10px;
 }
 
     @-webkit-keyframes room-fade
@@ -451,7 +453,7 @@ canvas {
 #userlist-lobby-owners li div.room-row,
 #owned-rooms li div.room-row {
     display: inline-block;
-    width: 240px;
+    width: 250px;
     padding: 5px;
     border: 1px solid #e5e5e5;
 }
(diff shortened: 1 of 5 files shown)


hnrkndrssn:
New layout

image

No note set
image

No note or flag set
image

JabbR/Content/themes/default/Chat.css
```diff
@@ -438,9 +448,10 @@ canvas {
             color: rgba(0, 16, 63, 1);
         }
 
-#userlist-lobby-owners li div.room-row {
+#userlist-lobby-owners li div.room-row,
+#owned-rooms li div.room-row {
     display: inline-block;
-    width: 250px;
+    width: 240px;
```
cmaish: This breaks locked and closed owned rooms in the lobby

cmaish:
Width issue aside, looks good to me.  Lists the user's rooms.

I don't think that we need to make them links at this stage - maybe in the distant future.  Linkifying them would either require getting a target for the main jabbr window so we could do <a href="https://jabbr.net/#/rooms/meta" target="mainjabbrwindowtarget">Meta</a> (which I don't think is possible with javascript); or to connect to the chathub (which I think is what we're probably gonna have to eventually do in order to get realtime updates to flags, notes and rooms).

hnrkndrssn:
Fixed the width issue on the Lobby

PR #1165, opened 2014-07-19
Add a create room button to the UI

### LINKED ISSUE #1154: Make room creation easier
People have trouble creating rooms.  We should add a button to the lobby that allows you to create one

hnrkndrssn:
The button will be visible on the lobby, and also when in a room (as you
might still want to create a new room while in another room)

In the lobby
image

In a room
image

Creating a new room
image

Without entering a room name
image

Entering a room name
image

image

Trying to create a room that already exists
image

Fix #1154.

### DIFF
--- a/JabbR/Nancy/HomeModule.cs
+++ b/JabbR/Nancy/HomeModule.cs
@@ -19,7 +19,7 @@ namespace JabbR.Nancy
 {
     public class HomeModule : JabbRModule
     {
-        private static readonly Regex clientSideResourceRegex = new Regex("^(Client_.*|Chat_.*|Content_.*|LoadingMessage)$");
+        private static readonly Regex clientSideResourceRegex = new Regex("^(Client_.*|Chat_.*|Content_.*|Create_.*|LoadingMessage|Room.*)$");
 
         public HomeModule(ApplicationSettings settings,
                           IJabbrConfiguration configuration,
@@ -41,7 +41,8 @@ namespace JabbR.Nancy
                         Version = Constants.JabbRVersion,
                         IsAdmin = Principal.HasClaim(JabbRClaimTypes.Admin),
                         ClientLanguageResources = BuildClientResources(),
-                        MaxMessageLength = settings.MaxMessageLength
+                        MaxMessageLength = settings.MaxMessageLength,
+                        AllowRoomCreation = settings.AllowRoomCreation
                     };
 
                     return View["index", viewModel];
--- a/JabbR/Nancy/HomeModule.cs
+++ b/JabbR/Nancy/HomeModule.cs
@@ -42,7 +42,7 @@ namespace JabbR.Nancy
                         IsAdmin = Principal.HasClaim(JabbRClaimTypes.Admin),
                         ClientLanguageResources = BuildClientResources(),
                         MaxMessageLength = settings.MaxMessageLength,
-                        AllowRoomCreation = settings.AllowRoomCreation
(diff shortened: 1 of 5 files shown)


cmaish:
The room topic covers the create room button in rooms so only the bottom 2px is clickable.  The topic needs to be shortened.
capture 38

I'm not convinced that creating a room is a common enough thing to show all the time - maybe we should get stats from jabbr.net on how many rooms are created in eg a month (having it on the lobby allows admins to not need to figure out how to use commands with no feedback though - saving the ~1 person each month who asks in meta how to create their first room [and also allows us to remove the message textbox from the lobby]).  I don't really mind either way, but if we have it, maybe we need to give it less prominence on the room pages.

When the screen is scaled down on the lobby, the create room button doesn't drop down to the next line:
capture 39

Not sure if it's in scope for this change (or whether we want to do it more generally), but we should also have a modal for when you create a room and receive an error when you're in the lobby, otherwise no error is shown and no room is created.  Maybe it needs more discussion, but there could be a convention where actions triggered via the UI (as opposed to entered as text into a room) using modals instead of writing to the current room.  As part of this, we'd shift to a modal prompt as well as a modal instead of alerts (similar to what the kick/ban stuff does now)

hnrkndrssn:
Will fix

hnrkndrssn:
Fixed layout issue when screen is scaled down and you're on the Lobby
image

Does not display the button when you're in a room
image

@davidfowl @cmaish Do we want to move this to using a modal dialog instead of using the default javascript prompt as suggested in the comments?

cmaish:
It's a step in the right direction and works on its own so I'd personally be happy with shipping it as-is and someone dialoging it up further later on.  Or if you are keen to keep running with it, feel free to do that too...

JabbR/Views/Home/index.cshtml
```diff
@@ -382,6 +382,12 @@
                     <input id="room-filter-closed" type="checkbox" />
                     @LanguageResources.Client_ShowClosedRooms
                 </label>
+                @if (Model.AllowRoomCreation)
```
davidfowl, line 385: Or is admin. Well leave the view model dumb and compute that in the controller

JabbR/Nancy/HomeModule.cs
```diff
@@ -41,7 +41,8 @@ public class HomeModule : JabbRModule
                         Version = Constants.JabbRVersion,
                         IsAdmin = Principal.HasClaim(JabbRClaimTypes.Admin),
                         ClientLanguageResources = BuildClientResources(),
-                        MaxMessageLength = settings.MaxMessageLength
+                        MaxMessageLength = settings.MaxMessageLength,
+                        AllowRoomCreation = settings.AllowRoomCreation
```
davidfowl: || IsAdmin

hnrkndrssn: Two options (I'm unsure which one would be better)

```
AllowRoomCreation = settings.AllowRoomCreation || Principal.HasClaim(JabbRClaimTypes.Admin)
```

or

```
viewModel.AllowRoomCreation = settings.AllowRoomCreation || viewModel.IsAdmin;
```

JabbR/Views/Home/index.cshtml
hnrkndrssn, line 385: So we keep this unchanged, right?

JabbR/Nancy/HomeModule.cs
cmaish: Maybe bikeshedding, but the first one looks better to me - it keeps it in the object initializer syntax and doesn't make assumptions on the other bits of the viewmodel.

PR #1189, opened 2014-08-05
Error Message when file upload is too large

### LINKED ISSUE #1190: Add a more descriptive error message when uploading a file that's over the size limit
When uploading a file that is over the file size limit, the following error message is displayed: 

`Failed to upload xxx.xxx`

The error message should be more descriptive...something like 

`Failed to upload xxx.xxx as it is larger than xMB`

### LINKED ISSUE #1191: Check for whitespace in name on client side when creating a room
chat.ui.js only checks for empty room name, and doesn't check if the name contains spaces

lagwagone:
Noticed a JavaScript error in IE9, which was caused by the call to
requestAnimationFrame in loader.js. Added client side checking of spaces on room name when creating a new room (fix #1191). Added custom error message for when a file upload is too large (fix #1190).


JabbR/loader.js
```diff
@@ -163,7 +163,7 @@ function run(arcs, ctx, canvas, frame, step) {
         arc = arcs[i];
         setupArc(arc, ctx, frame, speed);
     }
-    requestAnimationFrame(function () {
+    requestAnimFrame(function () {
```
cmaish, line 166: This is unrelated to the actual PR, but is a legit fix for an issue with IE9 (and possibly other browsers) where https://developer.mozilla.org/en/docs/Web/API/window.requestAnimationFrame isn't available.

JabbR/Chat.ui.js
```diff
@@ -1196,15 +1196,19 @@
...
                     msg = '/create ' + roomName;
                 if (roomName === null) {
                     return false;
                 }
                 else if (roomName === '') {
                     alert(utility.getLanguageResource('RoomNameCannotBeBlank'));
-                } else {
+                }
+                else if (/\s/.test(roomName)) {
+                    alert(utility.getLanguageResource('RoomNameCannotContainSpaces'));
```
cmaish, line 1209: Good catch.  We need something here to deal with invalid room name entries (anything other than [A-Za-z0-9_-]{1,30} ) when you're on the lobby screen - "a a is not a valid room name" might be what we wind up with for this eventually...

JabbR/Chat.ui.js
```diff
@@ -1196,15 +1196,19 @@
                 $downloadDialog.modal('hide');
             });
 
-            $createRoomButton.click(function () {
-                var roomName = prompt(utility.getLanguageResource('Create_CommandInfo')),
+            $createRoomButton.click(function () {              
+                var roomName = prompt(utility.getLanguageResource('Create_CommandInfo'), ''),
```
cmaish, line 1200: I'm getting the impression that you need to support IE8 - if this is the case, are you up for dropping into meta at some point to let us know if anything else is broken?

lagwagone, line 1200: I was running IE9 at work, but upgraded to IE11 on Friday so unfortunately can't continue to test the old version. The only issues that I noticed were with the JavaScript create room prompt defaulting to undefined and the JS error when calling requestAnimationFrame. I didn't spot any major issues while I was using the old browser.

PR #1201, opened 2014-09-12
Author positioning was poorly floating.

PureKrome:
Here's a screen shot of the NuGet content provider in JabbR (before the PR) ....

image

Noce the poor positioning of the author names? I tried this with the Json.NET package and it was doing the same thing.

So i just removed the `div` which had a `float` css key/value that seemed to be messing things up.

I just tested this with Chrome Dev Tools, playing around with the html/css values.

_Apologies: for not creating an Issue, first. Change seemed pretty small._

### DIFF
--- a/JabbR/Content/themes/default/Chat.nuget.css
+++ b/JabbR/Content/themes/default/Chat.nuget.css
@@ -118,10 +118,6 @@
     width: 50%;
 }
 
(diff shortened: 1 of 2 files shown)


davidfowl:
@PureKrome CI doesn't like this change.

PR #1213, opened 2014-12-22
Fixed deadlock race condition by locking access around non-threadsafe method ResourceManager.GetResourceSet

richardszalay:
We were running into a sporadic deadlock-on-startup issue in production (IIS 8 / Windows Server 2012 R2) and, after analysing the issue offline via WinDBG, traced it back to a race condition within ResourceManager.GetResourceSet as it was used in HomeModule.BuildClientResources(). 

The ResourceManager.GetResourceSet documentation definitely specifies it as "non threadsafe", but we have multiple localised LanguageResources which may explain why it hasn't been spotted on jabbr.net.

As a solution, we've added a lock (w/ cache) around access to it. We've not had the deadlock issue since we pushed the change.

### DIFF
--- a/JabbR/Nancy/HomeModule.cs
+++ b/JabbR/Nancy/HomeModule.cs
@@ -180,9 +180,30 @@ namespace JabbR.Nancy
             };
         }
 
-        private static string BuildClientResources()
+        private string BuildClientResources()
         {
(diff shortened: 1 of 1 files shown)


davidfowl:
Tests are failing!

richardszalay:
Unfortunately I don't have the time to dedicate to this anymore, so it's probably easier if I just close the PR.

Issue #1214, opened 2014-12-23
/giphy command


PureKrome:
Hi team!

before I start any work, i was wondering if I could add an `/giphy <keywords>` command.

[Giphy has an API](https://github.com/Giphy/GiphyAPI) which returns a url to a random image, tagged by the keywords you've provided :smile: Yep - moar pwitty pictures in #general-chat :gift: 

![Woot](http://i.giphy.com/bznNJlqAi4pBC.gif)

sample uri endpoint: http://api.giphy.com/v1/gifs/random?api_key=dc6zaTOxFJmzC&tag=yay

PR #1231, opened 2015-10-13
Giphy command implementation

timgranstrom:
Implemented the giphy command as described in issue #1214


JabbR/Commands/GiphyCommand.cs
```diff
@@ -0,0 +1,52 @@
...
+        public override void Execute(CommandContext context, CallerContext callerContext, ChatUser callingUser, string[] args)
+        {
+            if (String.IsNullOrEmpty(callerContext.RoomName))
+            {
+                throw new HubException(LanguageResources.InvokeFromRoomRequired);
+            }
+
+            ChatRoom callingRoom = context.Repository.GetRoomByName(callerContext.RoomName);
+
+            string url = String.Format("http://api.giphy.com/v1/gifs/random?api_key={0}&tag=", "dc6zaTOxFJmzC");
```
PureKrome, line 27: Should the API key be injected into this ctor or something? so then it can exist in the `<appSettings>` ... which are in azure so the data isn't public. ??

JabbR/Commands/GiphyCommand.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            //add keywords/tags to the url
+            int i = 0;
+            foreach (String keyword in args)
```
PureKrome, line 31: what about just doing something like ..

```
var tags = string.Join("&", args);
url = url + tags;
```

PureKrome:
@Buildstarted can you review this PR also, plz?

JabbR/Commands/GiphyCommand.cs
garysharp, line 31: Looks like [the API](https://github.com/Giphy/GiphyAPI#random-endpoint) doesn't support joining keywords in this way. it seems they need to be separated by a `space` character (which when encoded as [application/x-www-form-urlencoded](http://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.1) is represents as `+`).

Also consider sanitising the input:

``` C#
var tags = string.Join("+", args.Select(arg => Uri.EscapeDataString(arg)));
var url = string.Format("http://api.giphy.com/v1/gifs/random?api_key={0}&tag={1}", apiKey, tags);
```

PureKrome, line 31: LGTM!

JabbR/Infrastructure/Http.cs
```diff
@@ -58,5 +60,32 @@ public static Task<TResult> GetJsonAsync<TResult>(string url)
                 }
             });
         }
+
+        //Returns a webresponse as a string
+        public static string GetWebResponse(string url)
+        {
+
```
davidfowl, line 67: Fix the formatting here and make this async

JabbR/Infrastructure/Http.cs
```diff
@@ -58,5 +60,32 @@ public static Task<TResult> GetJsonAsync<TResult>(string url)
...
+                WebResponse response = request.GetResponse();
+                using (Stream responseStream = response.GetResponseStream())
+                {
+                    StreamReader reader = new StreamReader(responseStream, Encoding.UTF8);
+                    return reader.ReadToEnd();
+                }
+            }
+            catch (WebException ex)
+            {
+                WebResponse errorResponse = ex.Response;
```
davidfowl, line 80: This looks incomplete.

PR #1240, opened 2016-04-05
Correct server/client time offset

garysharp:
Server notifies client of time at hub connection and an offset is calculated. Any client created dates are corrected by the calculated offset. Possible solution to #950

> **Replaces #1041**
> _The original branch containing PR #1041 has long since been deleted, this new PR replaces it_

### DIFF
--- a/JabbR/Chat.js
+++ b/JabbR/Chat.js
@@ -1050,6 +1050,10 @@
         ui.updateTabOrder(tabOrder);
     };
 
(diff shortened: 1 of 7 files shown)


garysharp:
I've also incorporated PR #1042 (as 5fe47ac) for discussion as there are situations where calculating and applying the offset can still result in strange behaviour (namely with fuzzy times).

jabbrclienttimeerrormessage

